PR #121, opened 2016-07-27
improved collaborator data extraction scripts (MATLAB)

dimitri-yatsenko:


matlab/schemas/+preprocess/ExtractRaw.m
```diff
@@ -29,7 +28,7 @@
         
         function makeTuples(self, key)
             
-            %% extract masks and traces for Galvo Scanner
+            % extract masks and traces for Galvo Scanner
```
fabiansinz: Can we leave the %% in the future. They are useful to visually organize the code in the Matlab editor.

PR #122, opened 2016-07-27
Laser Calibration Table

fabiansinz:


python/pipeline/experiment.py
```diff
@@ -170,6 +172,60 @@ class Software(dj.Lookup):
         ('imager', '1.0')]
 
 @schema
+class LaserCalibration(dj.Manual):
+    definition = """
+    # stores measured values from the laser power calibration
+
+    calibration_id      : int
+    ---
+    -> Person
```
dimitri-yatsenko: I would add an explicit `calibration_date` attribute for those cases when the calibration date differs from the time that the data are entered in the database.

dimitri-yatsenko: Is it unreasonable to use a natural primary key for `LaserCalibration` e.g. (`->Rig, `calibration_date`)

dimitri-yatsenko: Why is software part of this table?

fabiansinz: Because it handles all the calibration that could influence the measurement (e.g. the Pockels cell).

fabiansinz:
I'll change the other things. Wait with acceptance.

PR #129, opened 2016-08-01
bugfix in population worker and key_source of ComputeTraces

fabiansinz:
- When the key_source of a computed relation was not properly defined, populate would do nothing but progress would still show that the relation is not fully populated. Because of that, the worker constantly tried to populate the remaining items, jamming the server. I added a time condition to account for that. 

- `preprocess.ComputeTraces`  had exactly that problem. Sometimes there were just no traces for a particular scan. I fixed that in the `key_source`.

### DIFF
--- a/python/scripts/worker-populate.py
+++ b/python/scripts/worker-populate.py
@@ -40,17 +40,9 @@ def main(argv):
 
     run_daemon = True # execute at least one loop
     while run_daemon:
-        populated_at_least_one = False
-        start = time.time()
         for name, rel in rels_cls.items():
-            todo, done = rel().progress()
-            if todo > 0:
-                populated_at_least_one = True
-                rel().populate(reserve_jobs=True)
-        end = time.time()
-        if run_daemon and not populated_at_least_one or (end - start) < 1:
-            t = np.random.randint(args.t_min, args.t_max)
-            time.sleep(t)
+            rel().populate(reserve_jobs=True, suppress_errors=True)
+        time.sleep(np.random.randint(args.t_min, args.t_max))
         run_daemon = args.daemon
     return 0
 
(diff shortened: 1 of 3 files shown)


python/scripts/worker-populate.py
```diff
@@ -41,13 +41,16 @@ def main(argv):
     run_daemon = True # execute at least one loop
     while run_daemon:
         populated_at_least_one = False
+        start = time.time()
         for name, rel in rels_cls.items():
             todo, done = rel().progress()
             if todo > 0:
                 populated_at_least_one = True
                 rel().populate(reserve_jobs=True)
-        if run_daemon and not populated_at_least_one:
```
dimitri-yatsenko: Why not always wait?  If it attempted to populate, then there is probably no need to check again immediately, right?

fabiansinz: Hmm, I guess you're right. I'll change it later.

dimitri-yatsenko: thanks. it should simplify the logic.

python/scripts/worker-populate.py
```diff
@@ -41,13 +41,16 @@ def main(argv):
     run_daemon = True # execute at least one loop
     while run_daemon:
         populated_at_least_one = False
+        start = time.time()
         for name, rel in rels_cls.items():
             todo, done = rel().progress()
```
dimitri-yatsenko: progress can be a time consuming operator and its function is duplicated in populate.  I would just run populate without checking progress.

python/scripts/worker-populate.py
```diff
@@ -41,13 +41,16 @@ def main(argv):
...
             if todo > 0:
                 populated_at_least_one = True
                 rel().populate(reserve_jobs=True)
```
dimitri-yatsenko: need to have the option to suppress errors.  populate does not catch errors by default.

PR #134, opened 2016-08-04
Now fetching HD5 filename from experiment.ScanWheelFile 

dimitri-yatsenko:


jakereimer:
I'm sympathetic with your distaste for "HD5file", and I appreciate adding the treadmill table to the db, but I would suggest "BehaviorFile" or "TreadmillFile" or "BallAndEyeFile" rather than ScanWheelFile. It makes it sound like we are scanning with a wheel like with a spinning disk confocal which is not accurate, and we typically refer to it as "treadmill" or "ball" in our manuscripts.

PR #139, opened 2016-08-11
 	add BehaviorSync and cleanup read_video_hdf5

fabiansinz:


python/pipeline/preprocess.py
```diff
@@ -627,6 +627,15 @@ class EyeQuality(dj.Lookup):
 
 
 @schema
+class BehaviorSync(dj.Imported):
+    definition = """
+    -> experiment.Scan
+    ---
+    frame_times                  : longblob # time stamp of imaging frame on behavior clock
```
dimitri-yatsenko: is behavior clock the same as stimulus clock?

python/pipeline/preprocess.py
```diff
@@ -872,64 +881,63 @@ def plot_traces(self, outdir='./'):
...
+        Needs opencv installation.
 
-def show_video(self, from_frame, to_frame):
-    """
-    Shows the video from from_frame to to_frame (1-based) and the corrsponding tracking results.
-    Needs opencv installation.
+        :param from_frame: start frame (1 based)
+        :param to_frame:  end frame (1 based)
+        """
+        if not len(self) == 1:
```
dimitri-yatsenko: len(self) != 1

python/pipeline/utils/eye_tracking.py
```diff
@@ -162,8 +83,6 @@ def ts2sec(ts, packet_length=0):
         not_zero = np.hstack((0, diffs + 1))
         f = iu_spline(not_zero, s[not_zero], k=1)
```
dimitri-yatsenko: So it's linear interpolation.  Why use spline interpolation?

PR #141, opened 2016-08-11
Added table to sync imaging frames to behavior

jakereimer:
Cleaned up `pipetools/ts2sec.m` and adjusted behavior so that it does not subtract timestamp offset (time of first sample). Subtracted offset in `preprocess.Sync` instead. 

`preprocess.BehaviorSync` gives times of imaging frames on behavior clock with no offset subtracted. `frame_times` can be used to align frames to treadmill and pupil traces


matlab/schemas/+preprocess/Sync.m
```diff
@@ -30,6 +30,8 @@ function makeTuples(self, key)
                 packetLen = dat.analogPacketLen;
             end
             datT = pipetools.ts2sec(dat.ts, packetLen);
+            datT = datT-datT(1); % subtract offset
```
fabiansinz: That's to define time zero as the start of the scan?

fabiansinz: Maybe this just demonstrates my ignorance of the sync again, but if you subtract `datT(1)` here but don't in `BehaviorSync`, how can the two timestamps ever match? They seem to refer to the same timestamps (both `dat.ts` loaded with a key from `Scan` with `preprocess.readHD5(key)`).

jakereimer: `datT` is only used to get the sampling rate of the behavior clock, so the offset doesn't matter. Removed the line where I subtracted the offset.

PR #145, opened 2016-08-16
Example script for pupil, trace, vis stim on same clock

jakereimer:
Example showing how to fetch and plot pupil x,y, and radius, synchronized with calcium data and visual stimulus info.

### DIFF
--- a/matlab/scripts/pupil_and_trace.m
+++ b/matlab/scripts/pupil_and_trace.m
@@ -1,0 +1,89 @@
+% In this script, we'll fetch the pupil radius and position, and plot it
+% along with a calcium trace, all on the behavior clock. The relative times
+% of the eye,treadmill, and trace are precise, but the clock itself starts
+% at some arbitrary offset.
+
+% Frame times from preprocess.Sync on the visual stimulus clock can be
+% used to synchronize behavioral variables with the visual stimulation.
+% We'll also plot the trial start times for this scan to demonstrate this.
+
+%% Choose a scan
+key.animal_id=8804;
+key.session=1;
+key.scan_idx=3;
+
+%% Eye
+% fetch pupil radius trace
+r = fetchn(preprocess.EyeTrackingFrame & key,'major_r');
+% ...then fetch pupil position traces
+center = fetchn(preprocess.EyeTrackingFrame & key,'center');
+
+% or fetch both together
+[r,center] = fetchn(preprocess.EyeTrackingFrame & key,'major_r','center');
+
+% undetected frames are nans in the radius trace
+detectedFrames = ~isnan(r);
+
+% convert positions from cells to array
+xy = nan(length(r),2);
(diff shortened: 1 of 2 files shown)


matlab/scripts/pupil_and_trace.m
```diff
@@ -0,0 +1,89 @@
...
+%% vis stim
+
+% fetch the frame times on the visual stimulus clock
+vt = fetch1(preprocess.Sync & key,'frame_times');
+vtSlice = vt(slice:nslices:end);
+
+% get the trials and for this scan and their flip times
+trials = fetch(vis.Trial * preprocess.Sync & key & 'trial_idx > first_trial and trial_idx < last_trial','flip_times');
+
+for i=1:length(trials)
```
dimitri-yatsenko: No need for i.  You can do
```
for trial in trials'
```

matlab/scripts/pupil_and_trace.m
```diff
@@ -0,0 +1,89 @@
...
+
+% Alternatively, you can fetch all the traces for this scan using fetchn
+%traces = fetchn(preprocess.ComputeTracesTrace & key,'trace');
+
+% Fetch the imaging frame times on the behavior clock and the number of slices per scan
+[ft, nslices] = fetch1(preprocess.BehaviorSync * preprocess.PrepareGalvo & key,'frame_times','nslices');
+
+% In a single scan with 3 slices, imaging frames are collected from slice 1, 2, 3, 1, 2, 3...
+% So there are nslices * length(tr) frame times
+assert(nslices*length(tr)==length(ft),'You should never see this message unless the scan was aborted')
```
dimitri-yatsenko: it turns out scans are cut short quite often.  So we need to handle this case.

matlab/scripts/pupil_and_trace.m
```diff
@@ -0,0 +1,89 @@
...
+% get pupil tracking times on the behavior clock
+et = fetch1(preprocess.Eye & key,'eye_time');
+
+% plot xy position and radius
+plot(et,r)
+hold on
+plot(et,xy)
+
+%% calcium traces
+% choose a random calcium trace
```
dimitri-yatsenko: random -> arbitrary

matlab/scripts/pupil_and_trace.m
jakereimer: I know but I really prefer to use a separate index since it makes debugging easier and you can use the index for other things if you need to

matlab/scripts/pupil_and_trace.m
jakereimer: Will do

matlab/scripts/pupil_and_trace.m
jakereimer: It's true that this can happen, but out of the 14 phase 1 mice this only occurred in two scans in mouse 9170. I would prefer that we address this problem when we get to it rather than including in this initial demo.

PR #148, opened 2016-08-23
Anatomy pipeline schema

olakiril:
This schema should be used for any aditional anatomical info that is not necessary for preprocessing


fabiansinz:
Can you add the respective python file as well. All you need to do is create a file `anatomy.py` in `python/pipeline` with the following content

```python
import datajoint as dj
schema = dj.schema('pipeline_anatomy') # change if this is not the database name

schema.spawn_missing_classes()
```

matlab/schemas/+anatomy/MaskCoordinates.m
```diff
@@ -0,0 +1,52 @@
...
+                'px_width','px_height','um_width','um_height');
+            depth = fetch1(experiment.Scan & key, 'depth');
+            
+            for imask = 1:length(mask_id);
+                
+                % get mask position
+                mask = masks{imask};
+                im = zeros(px_width,px_height);
+                im(mask) = weights{imask};
+                labeledImage = logical(true(size(im)));
```
dimitri-yatsenko: you don't need `logical()` here because `true` is already logical.

matlab/schemas/+anatomy/MaskCoordinates.m
```diff
@@ -0,0 +1,52 @@
...
+                measurements = regionprops(labeledImage, im, 'WeightedCentroid');
```
dimitri-yatsenko: looks like `labeledImage` does not actually get its pixels set to correct values.

olakiril:
@fabiansinz After talking with Dimitri, these should go under the preprocess
@dimitri-yatsenko first comment correct, second comment see how regionprops works. information is in im not labeledImage

PR #149, opened 2016-08-23
Added example script to identify dilating and constricting pupil periods

jakereimer:


matlab/scripts/compute_pupil_periods.m
```diff
@@ -0,0 +1,62 @@
...
+% low-pass filter below 1Hz
+x=rInterp;
+fs = 1/median(diff(pupilTime)); %Hz
+lowCut = 1;     %Hz
+k = hamming(2*round(fs/lowCut)+1);
+k = k/sum(k);
+l = size(x,1);
+n = length(k);
+n = floor(n/2);
+x = [x(n+2-(1:n),:); x; x(end-1-(1:n),:)];
```
dimitri-yatsenko: add comment that this mirror-reflects the boundaries to generate value outside

matlab/scripts/compute_pupil_periods.m
```diff
@@ -0,0 +1,62 @@
...
+key.scan_idx=3;
+
+%% fetch timestamps, fetch and filter pupil radius trace 
+r = fetchn(preprocess.EyeTrackingFrame & key,'major_r');
+pupilTime = fetch1(preprocess.Eye & key,'eye_time');
+
+% interpolate over undetected frames
+rInterp = interp1(pupilTime(~isnan(r)),r(~isnan(r)),pupilTime);
+
+% low-pass filter below 1Hz
```
dimitri-yatsenko: You can probably replace this with a one-line standard way to low-pass filter below 1 Hz.  We use the mirror reflection of boundaries and fftfilt to meet specific performance goals in calcium signals.

jakereimer:
I know it's a bit cumbersome but this is the filtering we did for the 2014 paper so I thought I would replicate that.

PR #153, opened 2016-08-25
Various functions

olakiril:
MaskCoordinates & ScanCoordinates extract information from the reader and should be in the preprocess structure. 

tuning RFs implement most things in matlab side


matlab/schemas/+anatomy/MaskCoordinates.m
```diff
@@ -0,0 +1,52 @@
...
+                'px_width','px_height','um_width','um_height');
+            depth = fetch1(experiment.Scan & key, 'depth');
+            
+            for imask = 1:length(mask_id);
+                
+                % get mask position
+                mask = masks{imask};
+                im = zeros(px_width,px_height);
+                im(mask) = weights{imask};
+                labeledImage = logical(true(size(im)));
```
dimitri-yatsenko: `logical` is unnecessary here.

matlab/schemas/+anatomy/MaskCoordinates.m
```diff
@@ -0,0 +1,52 @@
...
+                measurements = regionprops(labeledImage, im, 'WeightedCentroid');
```
dimitri-yatsenko: this does not seem like it should work since `labeledImage` is all true.

olakiril: It does since the mask is in im

dimitri-yatsenko: but that's not how it works, according to the documentation.  Have you tested this?  The mask is in the first image and the weights are in the second.

olakiril: It is how it works. the first argument provides the information about the different labeled regions, in which case it is all just one region, and the second argument is the image. The two have to have the same size.

PR #158, opened 2016-08-31
added require imageio to setup.py

jakereimer:


olakiril:
imageio library is not installed by default in the docker file and this requirement produces an error unless you install the imageio library every time you start a new docker. Please update the docker file.

fabiansinz:
That's what this pull request was about. I just checked. The new docker is build and contains `imageio`. Do a `docker pull ninai/pipeline` and you'll have it.

olakiril:
Yep it works, thanks

PR #159, opened 2016-08-31
population of Spikes and ComputeTraces works correctly in MATLAB

jakereimer:
Implemented copying of NMF spikes and copying of non-Twitch traces in MATLAB preprocess.Spikes and preprocess.ComputeTraces respectively.


fabiansinz:
Only request I have is to properly format the code (CTRL-a, CTRL-i).

PR #160, opened 2016-09-01
getTraces method for compansating time differences between slices

olakiril:


olakiril:
Also an fixed old MonetRF implementation that should not be used :)

jakereimer:
I think this should be put into a standalone function, not a method of Spikes since it's not sufficiently general. It gets time on the visual stimulus clock, but some scans don't have a visual stimulus. We encounter the same problem getting times on the behavior clock from BehaviorSync, so would have to duplicate the code for that. I do agree that we need some helper functions for timing and synchronization and I propose we start adding them to +pipetools.

olakiril:
Created a piptool helper function instead.

jakereimer:
Still think we may want a more general function eventually, but merging for now

PR #161, opened 2016-09-06
For piezo positive values are deeper in slice.

jakereimer:


olakiril:
Can you please double check?
I asked with you before I put that sign there, and the new position plot of the piezo shows + values go up.

PR #162, opened 2016-09-06
Correct piezo coordinates & nslices in AdjustedSpikes

olakiril:


jakereimer:
shouldn't `uslices == 1:nslices` always?

olakiril:
Not when you get single cells.

jakereimer:
You mean there may be a slice without ROIs? Ok - I can see that.

olakiril:
You can fetch the trace of just one cell or all the cells from one slice. uslices == 1:nslices  does not take that into account

PR #164, opened 2016-09-09
incorporated new fast_oopsi spike inference

dimitri-yatsenko:


python/pipeline/preprocess.py
```diff
@@ -4,6 +4,10 @@
 import numpy as np
 import sh
 import os
+try:
+    import pyfnnd
+except ImportError:
+    warn('Could not load pyfnnd.  Spike inference will fail. Clone oopsi from https://github.com/cajal/PyFNND.git')
```
fabiansinz: Can you be more specific? Only foopsi inference will fail. STM will still work.

requirements.txt
```diff
@@ -5,4 +5,5 @@ matplotlib
 seaborn
 sklearn
 h5py
+pyfnnd
```
fabiansinz: Can you add that to the `setup.py` as well?

dimitri-yatsenko: ok, just did

PR #170, opened 2016-09-27
Matlab MonetRF implementation

olakiril:


fabiansinz:
Why do we create double implementations of everything?  I don't think that's a good idea.

olakiril:
In general I agree, but the current python implementation is terribly inefficient. For example I am processing the MonetRF of one site since yesterday, and while it is using 500GB of memory it has not finished yet. 
Old matlab code was miles better, so if anything lets switch to that until Dimitri finishes his latest version.

fabiansinz:
Alright, but then raise an exception in the respective python code so that only one implementation is used at a time.

matlab/schemas/+tuning/MonetRF.m
```diff
@@ -72,9 +72,11 @@ function makeTuples(self, key)
                     update = conv2(movie, snippet(:,itrace)', 'valid')';
                     update = interp1((0:size(update,1)-1)/fps, update, (0:nbins-1)*bin_width);  % resample 
                     update = reshape(update', sz(1), sz(2), nbins);
+                    if any(isnan(update(:)));fprintf('nan values for trial # %d, skipping...',trial.trial_idx);break;end
```
dimitri-yatsenko: shouldn't this be `continue` rather than `break`?

PR #172, opened 2016-10-03
bugfix

jakereimer:


jakereimer:
because `toc` is used later in an `fprintf` statement and it throws error if `tic` has not yet been used without assignment.

jakereimer:
That's why we meg another tic

jakereimer:
It turns out that it was the first tic we didn't meg.

eywalker:
Looks like I suggested too late 😛

PR #181, opened 2016-10-21
Add table for new stimulus SingleDot

shenshan:


matlab/schemas/+vis/SingleDot.m
```diff
@@ -0,0 +1,18 @@
+%{
+vis.SingleDot (manual) # single dot to map receptive field
+-> vis.Condition
+-----
+luminance                   : float                         # cd/m^2
+contrast                    : float                         # Michelson's 0-1
```
dimitri-yatsenko: what is this the contrast of?  If you specify the brightness and contrast, then what do the `dot_color` and `bg_color` indicate?  Why is color 0-255?

olakiril:
Dimitri it seems that only can approve these changes, so please review the code.

PR #189, opened 2016-11-16
Fix issues with hdf5 file loading in new setup settings

jiakunf:


.gitignore
```diff
@@ -1,3 +1,8 @@
+# pipeline configuration
+.env
+pipeline_config.json
+.gitignore
```
eywalker: Why is `.gitignore` in the `.gitignore`?

PR #194, opened 2016-11-29
frame size preallocation bug fix

olakiril:


matlab/schemas/+preprocess/create_loader.m
```diff
@@ -47,7 +47,7 @@
 
 
 fprintf('\tLoading slice %d channel %d\n', islice, ichannel)
-scan = zeros(r,c, 1, length(frames));
+scan = zeros(c,r, 1, length(frames));
```
dimitri-yatsenko: this does not seem right.  In matlab rows are first, columns second.

olakiril: Correct but in preprocess.PrepareGalvo you get these sizes:
            sz = size(reader);
            key.px_height = sz(2);
            key.px_width  = sz(1);
You can rename the preparegalvo if you wish, but I have too many things populated like this and with my correction it works.

PR #198, opened 2016-12-08
galvo_corrections for motion and raster correction in python w/ changes

jiakunf:
galvo_corrections for motion and raster correction in python


eywalker:
In general, give more descriptive name to your pull request. In this case, your description "galvo_corrections for motion and raster correction in Python" would have been a good name.

python/pipeline/utils/galvo_corrections.py
```diff
@@ -0,0 +1,61 @@
...
+                                                                                        raster_phase)/fill_fraction)
+    return img
+
+def plot_raster():
+
+    import matplotlib.pyplot as plt
+    import seaborn as sns
+    from pipeline import preprocess
+    from tiffreader import TIFFReader
+    reader = TIFFReader('/Users/titan/data/cache/11676_4_00006_00009.tif')
```
eywalker: Usually it's not a good idea to have a hard coded file path like this in a repository intended to run on multiple platforms. Consider moving the file path as an argument to this function `plot_raster`.

python/pipeline/utils/galvo_corrections.py
```diff
@@ -0,0 +1,61 @@
+from scipy.interpolate import interp1d, interp2d
+import numpy as np
+
+
+def correct_motion(img, xymotion):
```
eywalker: For this function (`correct_motion`) and `correct_raster`, it would be really useful to have some doc string explaining what these two functions do. Please add some simple documentation.

python/pipeline/preprocess.py
```diff
@@ -88,6 +89,13 @@ class Galvo(dj.Part):
         raster_phase            : float             # shift of odd vs even raster lines
         """
 
+        def get_fix_raster(self):
```
eywalker: Please add a simple docstring explaining what this does. Something along the line of `returns a function that performs raster correction on the image`.

python/tests/test_preprocess.py
```diff
@@ -0,0 +1,38 @@
...
+
+def test_rastercorrect():
+    """test raster correction"""
+
+    # img = np.random.randn(512, 512, 1, 1, 100)
+    # raster_phase = 0.1
+    # fill_fraction = 1
+    from pipeline import preprocess
+    from tiffreader import TIFFReader
+    reader = TIFFReader('/Users/titan/data/cache/11676_4_00006_00009.tif')
```
eywalker: Same as above - it's not a good idea to tie your tests to a specific machine. For the test purpose, you can consider moving the `.tif` image directly into the `tests` directory if it's not too large.

eywalker:
Changes look good. Thanks for incorporating them.

PR #199, opened 2016-12-19
bugfix in PrepareGalvo

fabiansinz:
- ratio was the wrong way around. If zoom=1 and the closest zoom is 1.1, then the scaling for the FOV should be 1.1/1 not 1/1.1.


eywalker:
Thanks for finding this and figuring out the bug.

olakiril:
Found this independently today, what are the chances. I also accounted for different aspect ratios

PR #201, opened 2016-12-19
Correct FoV calculation

olakiril:
Fixed a bug + accounted for multiple aspect ratios


olakiril:
Yes, both 5.1 & 5.2 work.
We have not used anything older.

PR #203, opened 2017-01-16
New stimulus, segmentation ROI matching, CaImAn, cleanup

fabiansinz:
- created new stimulus set called MadAlex
- cleanup in ipython notebooks and threw out obsolete files
- updated dockerfile in preparation to move nmf (caiman) to python
- add table to match segmentation ROIs between scans

Sorry for the monster request.


python/pipeline/preprocess.py
```diff
@@ -13,10 +14,15 @@
...
 from .utils import galvo_corrections
+from .experiment import Session, Scan
+import imreg_dft as ird
+
+# import caiman.source_extraction.cnmf as cnmf
 
 assert StrictVersion(dj.__version__) >= StrictVersion('0.2.8')
 
 schema = dj.schema('pipeline_preprocess', locals())
+schema.spawn_missing_classes()
```
dimitri-yatsenko: `spawn_missing_classes` should be called at the end of the module

python/pipeline/preprocess.py
```diff
@@ -624,7 +631,11 @@ def _make_tuples(self, key):
         if method == 'stm':
             prep = (Prepare() * Prepare.Aod() & key) or (Prepare() * Prepare.Galvo() & key)
             fps = prep.fetch1['fps']
-            X = (ComputeTraces.Trace() & key).proj('trace').fetch.as_dict()
+            X = []
```
dimitri-yatsenko: this is asking for a list comprehension

python/pipeline/preprocess.py
```diff
@@ -185,8 +190,11 @@ class ExtractRaw(dj.Imported):
 
     @property
     def key_source(self):
-        return Prepare() * Method().proj() & dj.OrList(
-            Prepare.Aod() * Method.Aod(), Prepare.Galvo() * Method.Galvo())
+        return Prepare() * Method() \
+               & ((Prepare.Galvo() * Method.Galvo() - 'segmentation="manual"') | \
```
dimitri-yatsenko: the __or__ operator is not implemented.  What is meant here?  Is this copied from MATLAB?  Use `dj.OrList` instead.

python/pipeline/preprocess.py
fabiansinz: Why? We want to load all classes and then replace those for which we have actual implementation. Otherwise we'll overwrite the existing classes.

python/pipeline/preprocess.py
fabiansinz: Yeah, it is. I'll update that .

python/pipeline/preprocess.py
fabiansinz: Ok, I just check that `spawn_missing_classes` checks for existing classes. It does. Moved it to the end.

PR #207, opened 2017-01-26
Object for cortical areas mapping

olakiril:


matlab/schemas/+map/getSchema.m
```diff
@@ -0,0 +1,7 @@
+function obj = getSchema
+persistent schemaObject
+if isempty(schemaObject)
+    schemaObject = dj.Schema(dj.conn, 'map', 'manolis_map');
```
fabiansinz: that should become `pipeline_map`

PR #209, opened 2017-01-30
Create `EnterProcessing` as gate into pipeline

fabiansinz:


olakiril:
What is the difference between having an "EnterProcessing" vs "ScanIgnored".
With the online preprocessing we can have a set of conditions that if not met a flag will be raised. Right now ScanIgnored is automatically populated from the Labview program once you click "cancel scan", and it should be fairly easy to add mean fluorescence & movement thresholds.

fabiansinz:
`ScanIgnored` and `EnterProcessing` does basically the same thing, but from different angles. 
I think using a positive list we'll end up analyzing less trash. Let's say you scanned and you come back and the end to find out that something has failed. You program has not entered something into the `ScanIgnored` table. So people have to actively think about that, which they won't. If we analyze scans for trials like that we'll have a lot of entries in the jobs table which fail which have to be dealt with. 

Furthermore, when we run experiments on other setups, we won't necessarily have your program. 

Wouldn't it be easy to have a dialog at the end of the scanning process in your program that says "Do you want to forward this scan to processing?". That way we achieve the same thing but with less rubbish and error that have to be dealt with.

olakiril:
All these things should be automated. That was the reasoning behind the Labview program. We will only put and remove mice from the setup, and everything else will be done programmatically so there can be no pop up windows. As it is now, if someone does not want to enter the scan into the database, they can click Cancel before the scan ends or manually enter that scan into the ScanIgnored table. 
If we go the "EnterProcessing" route, I can have the program enter it automatically unless the scan is canceled or some parameter is not met, but then we will have the same problem with just more tuples to deal with.

fabiansinz:
Neither of your proposals prevents processing of rubbish. I would like to have one interaction with the user where (s)he decides to keep the scan or not. Everything else should not be processed. I agree that everything should be automated as much as possible. But the current way produces too many false positives. Then we have to deal with the errors later which cannot be automated. So it just shifts the problem instead of solving it.

olakiril:
In case something awry happens during a scan and it doesn't fit the pipeline criteria the user should cancel it. Waiting till the end makes no sense unless you want to process it in which case it should go through the pipeline anyways. If again the user is not there during the scan to detect any problems (most common case), then a dialog at the end will not make any difference since you will not know if something went bad.

fabiansinz:
Check `preprocess.Prepare().popRel & 'animal_id=12486'` which contains 16 scans. According to Paul at most 3 of them are good. This means in the current setting I populate the database with 13 scans that are likely never used. That wastes space and resources. `EnterProcessing` would have prevented this, `ScanIgnored` not. I am open to anything that solves exactly that problem.

olakiril:
Again, the user should use the "Cancel" button if a scan should not be populated. Maybe I wasn't clear to Paul when I told him that. Also scans that are "likely never used" is quite broad. From my understanding Paul was doing RF mapping for many of these sites, in which case you need the data in the pipeline.

fabiansinz:
Even if there are RF mapping scanns we are probably still at more than 10 scans that nobody uses. And the argument "the user shouldn't have done that" is not a good one because the user provably did do those things which will happen again. I don't understand your resistance against a positive list. In one case the user has to click "cancel" in the other the user has to click "please process". The number of user interactions are exactly the same only the default changes because the current default, although a nice idea, has proven to be not robust.

olakiril:
Fabian, I believe we both agree that there is no difference between having a ScanIgnored & a EnterPipeline table, since you can get the same information from both cases. If anything you will end up with more tuples to deal with in the second case. Also since there are a lot of dependencies on the Labview side that are not dealt with in this pull request, it will actually break things and cannot be implemented as is. 
What you ask for is a user interaction change that has nothing to do with the pipeline code.

fabiansinz:
So how to solve it? I basically cannot proceed switching on automatic processing without resolving that issue because that will flood the database with unnecessary data and the `jobs` table with a lot of error messages.

olakiril:
Again, renaming a table will not solve any problem. If there are error messages then we should figure out how to handle the errors, because ideally if the scan is fine it should not cause any. What are the errors that you had with the recent recording from Paul?

PR #214, opened 2017-02-03
fix a)Improved the code of OptImageBar b)DotRF supports piezo c)NMF spikes only with NMF extraction

olakiril:


matlab/schemas/+map/OptImageBar.m
```diff
@@ -70,17 +69,20 @@ function makeTuples( obj, key )
...
+                    try
+                        disp 'getting the vessels...'
+                        k = [];
+                        k.session = key.session;
+                        k.animal_id = key.animal_id;
+                        k.site_number = fetch1(experiment.Scan & key,'site_number');
+                        ves_key = fetch(experiment.Scan & k & 'aim = "vessels"');
+                        reader = preprocess.getGalvoReader(ves_key);
+                        vessels = reader(:,:,:,:,:);
+                        vessels = mean(vessels(:,:,:),3);
```
dimitri-yatsenko: What is being done here? Is vessels 5-dimensional or 3-dimensional?

PR #219, opened 2017-02-21
New dockerfile to deal with new requirements and script to show uncorrected vs corrected images

ecobost:


python/pipeline/utils/show_galvo_correction_results.py
```diff
@@ -0,0 +1,77 @@
+# Written by: Erick Cobos
```
fabiansinz: Can you actually make that a member function of `preprocess.Prepare.Galvo`? The function should take the key as argument and the filename where to store the video and then save the video there.

PR #221, opened 2017-02-27
Couple of minor changes in show_video and merging them with the master

ecobost:


python/pipeline/preprocess.py
```diff
@@ -196,11 +196,16 @@ def save_video(self, filename='galvo_corrections.mp4', slice=1, channel=1,
 
         fig = plt.figure()
         plt.subplot(1, 2, 1);
-        plt.title('Original')
-        im1 = plt.imshow(original_scan[:, :, 0])  # just a placeholder
+        plt.title('Original');
```
fabiansinz: Please remove semicolon

Dockerfile
```diff
@@ -67,7 +67,7 @@ RUN apt-get update -y -q && \
     apt-get install -y clang-3.9 lldb-3.9 && \
     apt-get install -y libc6-i386  libsuitesparse-dev && \
     export LLVM_CONFIG=/usr/lib/llvm-3.9/bin/llvm-config && \ 
-    git clone --recursive https://github.com/simonsfoundation/CaImAn.git && \
+    git clone --recursive https://github.com/ecobost/CaImAn.git && \
```
fabiansinz: Don't you want to make a PR to the original one and we'll use that? The problem with using your version is, that it's harder to get updates.

python/pipeline/preprocess.py
```diff
@@ -504,6 +511,14 @@ def _make_tuples(self, key):
         # Save traces in their apropiate schemas.
         raise NotImplementedError('ExtractRaw is populated in Matlab')
 
+    def cnmf_save_video(self):
+        # Acces location_matrix, activity_matrix, and all that's needed
+        # Copy from caiman_interface save video.
+        pass
```
fabiansinz: Insert blank line or let pycharm format the code for PEP8.

Dockerfile
ecobost: I did but I don't know whether they'll merge it soon or solve it in a different way. We can let the original for now if you want and hope they do merge it or I'll wait until they send us the dendrite version and then fork and change that version

ecobost:
I thought my previous pull request was already merged. I still need to do check some things in the code that i pushed friday, I'll do it early tomorrow

Dockerfile
fabiansinz: No, then let's use your solution for now and change it when they merge it.

PR #225, opened 2017-03-14
Implementation of ExtractRaw in python

ecobost:
Implements the make tuples of ExtractRaw (and helper functions). Defines new dj.Part classes in ExtractRaw (GalvoCorrelationImage, BackgroundComponents, AR Coefficients and CNMFParameters) to save the configuration params and results of CNMF (other things are saved in Trace, GalvoROI and SpikeTrace in the same format they were saved in the matlab version: F-ordered, index starting at 1, etc).


python/pipeline/experiment.py
```diff
@@ -383,6 +384,16 @@ class Laser(dj.Part):
         gdd: float  # gdd setting
         """
 
+    def get_local_filename(self):
```
ecobost: get_local_tif_wildcard does not sound quite right to me , i put local_filename_with_tif_wildcard, it's very long but i couldn't think of something better

PR #229, opened 2017-03-30
ExtractRaw make_tuples working for soma and axons

ecobost:


Dockerfile
```diff
@@ -77,11 +77,11 @@ RUN apt-get update -y -q && \
     apt-get install -y clang-3.9 lldb-3.9 && \
     apt-get install -y libc6-i386  libsuitesparse-dev && \
     export LLVM_CONFIG=/usr/lib/llvm-3.9/bin/llvm-config && \ 
-    git clone --recursive https://github.com/simonsfoundation/CaImAn.git && \
+    git clone --recursive https://github.com/ecobost/CaImAn.git && \
```
dimitri-yatsenko: this should probably be forked to `github.com/cajal` or `/atlab` and cloned from there.

python/pipeline/preprocess.py
```diff
@@ -130,6 +130,21 @@ def estimate_num_components_per_slice(self):
 
             return int(round(num_components))
 
+        def estimate_neuron_size_in_pixels(self):
+            """ Estimates the size of a neuron in the scan (in pixels). Assumes neurons
+             are 15 x 15 microns."""
```
dimitri-yatsenko: typical mouse neurons are 7-9 microns in diameter.

python/pipeline/preprocess.py
```diff
@@ -130,6 +130,21 @@ def estimate_num_components_per_slice(self):
...
+            neuron_size_in_microns = 15 # assumption
+
+            # Calculate size in pixels
+            um_width, px_width = (Prepare.Galvo() & self).fetch1['um_width', 'px_width']
+            one_micron_in_pixels = px_width / um_width
```
dimitri-yatsenko: There is a possibility of deviations from isometry of pixels:  `um_width/px_width` may differ from `um_height/px_height`.  Perhaps we should at least check that the deviation is not serious or handle it somehow if it is.

dimitri-yatsenko: The inverse of `one_micron_in_pixels`, i.e. `microns_per_pixel` aka `pixel pitch` is commonly used in imaging systems and is more expressive.

python/pipeline/preprocess.py
```diff
@@ -130,6 +130,21 @@ def estimate_num_components_per_slice(self):
...
+            neuron_size_in_pixels = neuron_size_in_microns * one_micron_in_pixels
+
+            # Make sure it is int and at least 1.
+            neuron_size_in_pixels = int(max(round(neuron_size_in_pixels), 1))
```
dimitri-yatsenko: Why round and clamp it here?  The doc string does not mention that these transformations.  I suggest returning the actual size in pixels and rounding and clamping in the calling context.

python/pipeline/preprocess.py
```diff
@@ -130,6 +130,21 @@ def estimate_num_components_per_slice(self):
...
+
+            return neuron_size_in_pixels
```
dimitri-yatsenko: `size` is not as precise as `diameter` or `radius`.  I suggest using `diameter`.

python/pipeline/preprocess.py
```diff
@@ -779,24 +815,31 @@ def update_img(i):
...
         return fig
 
-    def plot_contours(self, slice=1, channel=1):
-        """ Draw contours of all masks over the correlation image.
+    def plot_contours(self, slice=1, channel=1, first_n=None):
+        """ Draw contours of masks over the correlation image.
 
         :param slice: Scan slice to use
         :param channel: Scan channel to use
+        :param first_n: Number of masks to plot. None for all.
```
dimitri-yatsenko: "None or 0 for all."

Dockerfile
ecobost: I plan to change it back to simonsfoundation when they accept my PR

python/pipeline/preprocess.py
ecobost: Set None as default. I think is better than 0 which means something, it could be -1 but I don't see why it would be better

python/pipeline/preprocess.py
ecobost: I rounded it because I wasn't thinking of subpixel sizes (10.2 pixels or 0.4 pixels) but I guess is possible, I'll change it.

Dockerfile
dimitri-yatsenko: better yet, you can fork it under `github.com/cajal` and then we can accept the changes when we are ready and not wait for them.

dimitri-yatsenko: The problem with keeping it under `ecobost` is that our production code will potentially be relying on your development code that has not been reviewed.

python/pipeline/preprocess.py
dimitri-yatsenko: At the risk of being pedantic, I think you should refer to the `soma` or `cell body` since most of the neuron is not in the soma.

ecobost: I changed names in the new version to estimate_soma_radius_in_pixels and now it returns a tuple of radius(es?), I left the estimate of the neuron at 14 (thus radius at 7) because if I put less it says neuron are smaller than they actually are in the scan (tested in three different scans)

python/pipeline/preprocess.py
dimitri-yatsenko: Then the condition should be `if first_n is not None:`

PR #230, opened 2017-04-06
Stimuli in +stimulus 

dimitri-yatsenko:
Working stimuli:
* stimulus.Varma
* stimulus.Matisse
* stimulus.Matisse2
* stimulus.Clip


matlab/schemas/+stimulus/Clip.m
```diff
@@ -0,0 +1,61 @@
...
+-----
+-> stimulus.MovieClip
+cut_after              : float           # (s) cuts off after this duration
+%}
+
+classdef Clip < dj.Manual & stimulus.core.Visual
+    
+    properties(Constant)
+        variation = '1'
+        movie_dir = '/Users/dimitri/stimuli'
```
ecobost: Should there be a different default here?

matlab/schemas/+stimulus/Matisse2.m
```diff
@@ -0,0 +1,159 @@
...
+                self.flip(opts)
+                opts.checkDroppedFrames = true;
+                Screen('close',tex)
+            end
+        end
+    end
+end
+
+
+function img = upscale(img, factor, ori, coherence, nframes, speed)
```
ecobost: These functions could be indexed. It looks nicer

dimitri-yatsenko: Did you mean "indented"?  I use MATLAB's indentation rules, which are enforced when you mark all text (Ctrl-A) and then press Ctrl-I.   These are function that are external to the class and they are not indented in MATLAB.

matlab/schemas/+stimulus/Clip.m
dimitri-yatsenko: Yes, it should be `~/stimuli` but PsychToolbox does not understand `~`.  we should ask Manolis what it this should be.

ecobost: In +stimulus/MovieClip.m you put movie_dir='movies', maybe movie_dir='stimuli' is good here

PR #234, opened 2017-04-19
Updated config file for Varma stimulus

rajdbz11:
Updated the config file for Varma stimulus:
Stimulus sequence: 
S and L correpond to one major segment for short and long range correlation conditions in the Varma stimulus (10 mins long each)
M is Monet (30 mins long)
The sequence we want is S, L, S, L, M, S, L, S, L  (Total duration = 110 mins)

Also, within each S and L segment we have we have 5 repetitions each of short 's' and 'l' clips randomly interleaved.


matlab/schemas/+stimulus/+conf/Varma.m
```diff
@@ -2,38 +2,123 @@
 control = stimulus.getControl;
 control.clearAll()   % clear trial queue and cached conditions.
 
+paramsM.fps             = 60;
+paramsM.duration        = 180;  % to be changed to 1800s
```
dimitri-yatsenko: I recommend against making very long trials.  In our current framework, we do not log partially completed trials.  So if for some reason we need to interrupt the experiment, the ongoing trial is lost.  Therefore, the longest trials we have are ~60 seconds in long experiments.

Of course, we could change the framework to allow processing partially completed trials, increasing the amount of programming and testing required to handle special cases such as interrupted trials.

matlab/schemas/+stimulus/+conf/Varma.m
```diff
@@ -2,38 +2,123 @@
...
+paramsM.rng_seed        = 1;
+paramsM.pattern_width   = 32;
+paramsM.pattern_aspect  = 1.7;
+paramsM.pattern_upscale = 10;
+paramsM.ori_coherence   = 2.5;
```
dimitri-yatsenko: The Monet2 parameters have changes.  Please update as in stimulus.conf.matisse2.

matlab/schemas/+stimulus/Varma.m
```diff
@@ -59,23 +59,26 @@ function test()
             rng(cond.noise_seed)
```
dimitri-yatsenko: the `rng` should be used inside the `make` method so that it will work when called by the stimulus framework.

matlab/schemas/+stimulus/Varma.m
```diff
@@ -206,9 +209,9 @@ function test()
                 YsVid(:,:,tt) = upscale(Y(:,:,tt), fmask, f_up);
             end
             
-            Tstart  = 1 + cond.fps*cond.pre_blank_period;
+            Tstart  = 1 + cond.fps*init_buffer_period;
             YsVid   = YsVid(:,:,Tstart:end);         
-            K       = 0.03;
+            K       = 0.025;
             cond.movie = uint8(round(256*(YsVid/2/K + 0.5)));
```
dimitri-yatsenko: 256 -> 255

PR #236, opened 2017-04-21
Created separate config file VarmaTest

rajdbz11:
Two config files for Varma:
VarmaTest (11 mins)
and
Varma (110 mins )


matlab/schemas/+stimulus/+conf/Varma.m
```diff
@@ -4,7 +4,7 @@
 
 cond.fps = 60;
 cond.duration = 30;
-cond.rng_seed = 1:6; % to be changed to 1:60 to generate Monet stim for 1800 s
+cond.rng_seed = 1:60; % to be changed to 1:60 to generate Monet stim for 1800 s
```
dimitri-yatsenko: unnecessary comment

PR #251, opened 2017-06-22
Added nmf as a deconvolution method, fixed check for display (so it runs if DISPLAY is not set)

ecobost:


python/pipeline/reso.py
```diff
@@ -514,10 +511,6 @@ def _make_tuples(self, key):
             self.notify({**key, 'slice': slice_id + 1}, scan) # once per slice
 
     def notify(self, key, scan):
-        # --  notification
-        import matplotlib
-        matplotlib.rcParams['backend'] = 'Agg'
-        import matplotlib.pyplot as plt
```
fabiansinz: Is `plt` imported somewhere else in the file? Setting it to 'Agg' is ok to do at the top, but I am not sure whether you can get rid of the `import matplotlib.pyplot as plt` import.

ecobost: Yep, I tested it. pyplot is imported at the top so setting the backend here has no effect, the other option is to import it at every function call but then plt could be imported in some other place or in the same python session and thus setting the backend in the specific function would have no effect. I fixed it so when pipeline is imported the correct backend is set if there is or isn't a display to show, so if you don't have a display, it still works and if you have a display a window does open when creating the notifications but closes almost instantly, maybe there is an option to call plt.figure with the default backend and not plot

fabiansinz: Sorry, what I meant is: I couldn't find a place where `plt` is now imported. Is it still available? Did you test whether it correctly plots and sends the images?

ecobost: It is imported at the very top of reso.py (because it is used in a lot of places). I did test it, I'll check again now while I try to rescale cnmf traces to the right range(I think I know how, so don't merge this yet)

PR #255, opened 2017-06-27
Fixing bug when plotting single slie and single channel (for slack notifications)

ecobost:
Some other minor changes to how notifications are plotted


ecobost:
New PR with all changes

PR #256, opened 2017-06-28
Meso pipeline

ecobost:
Steps are pretty similar to those in reso, except that some information has been added (such as height width, um_x, um_y and rois per field) some other deleted (such as zoom) and each mask has a delay in milliseconds from the start of the frame (in ScanSet.UnitInfo). In theory they could now keep diverging as we add more scan-specific steps (for instance if we want to do some tiling in mesoscope).
Here are the steps:
```python
from pipeline import meso
key = {'animal_id': 14445, 'session': 3, 'scan_idx': 3}

meso.ScanInfo().populate(key) # read header info, compute quantal size and compute delays per field

# manually choose channel for correction (notice field)
meso.CorrectionChannel().insert1({**key, 'field': 1, 'channel': 1})
meso.CorrectionChannel().insert1({**key, 'field': 2, 'channel': 1})
meso.CorrectionChannel().insert1({**key, 'field': 3, 'channel': 1})
meso.CorrectionChannel().insert1({**key, 'field': 4, 'channel': 1})

meso.RasterCorrection().populate(key) # raster correction, starts iff all fields are in CorrectionChannel
meso.MotionCorrection().populate(key) # motion correction
meso.SummaryImages().populate(key) # average and correlation image

# manually choose what to segment (segmentation_method 1=manual, 2=nmf)
meso.SegmentationTask().insert1({**key, 'field': 1, 'channel': 1, 'segmentation_method': 2, 'compartment': 'soma'})
meso.SegmentationTask().insert1({**key, 'field': 2, 'channel': 1, 'segmentation_method': 2, 'compartment': 'soma'})
meso.SegmentationTask().insert1({**key, 'field': 1, 'channel': 1, 'segmentation_method': 1, 'compartment': 'soma'})

meso.Segmentation().populate(key) # nmf or manual (in Matlab) segmentation to create masks
meso.Fluorescence().populate(key) # creates calcium traces for manual masks (in Calcium.Trace), nmf populates this directly
# meso.MaskClassification().populate(key) # opens a GUI to manually classify each mask (manual segmentation populates this directly)
meso.ScanSet().populate(key) # gives each mask a unique id per scan (unit_id) and creates UnitInfo with type, coordinates and ms_delay
meso.Activity().populate({**key, 'spike_method': 5}) # deconvolves spike traces with different methods {3: stm, 2: oopsie and 5: nmf)
```
There is also some small improvements and fixes in reso

### DIFF
--- a/python/pipeline/utils/eye_tracking.py
+++ b/python/pipeline/utils/eye_tracking.py
@@ -1,13 +1,9 @@
 from collections import defaultdict
-
-import matplotlib
+import matplotlib.pyplot as plt
 import pandas as pd
+import numpy as np
 
 from ..exceptions import PipelineException
-
-matplotlib.use('agg')
-import matplotlib.pyplot as plt
-import numpy as np
 
 try:
     import cv2
--- a/python/pipeline/__init__.py
+++ b/python/pipeline/__init__.py
@@ -7,7 +7,7 @@ _report_on = {
 
 
 #--- switch matplotlib backend if there is no way to display things.
-cmd = 'python -c "import matplotlib.pyplot as plt; plt.figure()" 2> /dev/null'
+cmd = 'python3 -c "import matplotlib.pyplot as plt; plt.figure()" 2> /dev/null'
 if os.system(cmd): # if command fails
     print('No display found. Switching matplotlib backend to "Agg"')
     import matplotlib; matplotlib.use('Agg'); del matplotlib
--- a/python/pipeline/behavior.py
+++ b/python/pipeline/behavior.py
@@ -1,12 +1,9 @@
-import os
-
 import datajoint as dj
 from .utils.signal import spaced_max, longest_contiguous_block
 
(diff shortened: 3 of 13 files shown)


python/pipeline/utils/eye_tracking.py
```diff
@@ -1,14 +1,10 @@
 from collections import defaultdict
-
-import matplotlib
+import matplotlib.pyplot as plt
 import pandas as pd
+import numpy as np
 
 from ..exceptions import PipelineException
 
-matplotlib.use('agg')
```
ecobost: @fabiansinz fabian, is this ok?, I deleted it because it will print a warning every time we import something from utils (because eye_tracking is imported in utils.__init__). If not, maybe I can add it back and pass keyword argument warn=False or just have it print the warning

fabiansinz: I think it's fine. Your solution to the matplotlib problem should solve that as well.

PR #260, opened 2017-07-02
Added ScanDone at end of reso/meso for scan-level processing

ecobost:
Classes that can process pipeline outputs (masks, traces, spikes, info) per slice can depend on Activity (key is scan, slice, channel), classes that need to process the scan as an atomic object can depend on ScanDone (key is scan), which takes cares of keeping itself updated. This way processing in the pipeline can be done per slice but results in further steps could be processed per slice as they come or per scan in a bundle.
Also, changed dependency of ScanSet from Segmentation to Fluorescence (it should have been like that but I didn't notice before)

### DIFF
--- a/python/pipeline/meso.py
+++ b/python/pipeline/meso.py
@@ -1370,7 +1370,7 @@ class ScanDone(dj.Computed):
         # Reinsert in ScanDone
         self.insert1(scan_key)
 
-        # Insert all processed slices in Partial
+        # Insert all processed fields in Partial
         ScanDone.Partial().insert((Activity() & scan_key).proj())
 
         self.notify(scan_key)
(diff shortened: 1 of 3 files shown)


python/pipeline/meso.py
```diff
@@ -1232,7 +1243,7 @@ def _make_tuples(self, key):
 
         # Get fluorescence
         fps = (ScanInfo() & key).fetch1('fps')
-        unit_ids, traces = ((ScanSet.Unit() & key) * Fluorescence.Trace()).fetch('unit_id', 'trace')
+        unit_ids, traces = (Fluorescence.Trace() * (ScanSet.Unit() & key)).fetch('unit_id', 'trace')
```
ecobost: I actually thought it was the other way around, I'll change it

PR #265, opened 2017-07-19
Removed the stimulus schema from the public pipeline

dimitri-yatsenko:
* Added `reso` and `shared` schemas to the matlab pipeline.


python/pipeline/reso.py
```diff
@@ -1065,13 +1065,20 @@ def _make_tuples(self, key):
...
+
+
 @schema
 class ScanSet(dj.Computed):
     definition = """ # set of all units in the same scan
-
     -> Fluorescence                 # processing done per slice
     """
 
+    def _job_key(self, key):
```
ecobost: Will this be called job_key or _job_key, because we already have job_key below. I don't think any of them work now so we can delete it later

PR #279, opened 2017-10-23
Added the fuse schema to merge the two-photon pipelines

dimitri-yatsenko:


python/pipeline/fuse.py
```diff
@@ -0,0 +1,64 @@
...
+
+    class Meso(dj.Part):
+        definition = """
+        -> meso.ScanDone
+        -> ScanDone
+        """
+
+    def _make_tuples(self, key):
+
+        for pipe, src, dest in (
```
ecobost: Could we do this check with ifs, the for else clause is a bit confusing:
```python
if reso.ScanDone() & key:
   pipe, src, dest = 'reso', reso.ScanDone, ScanDone.Reso
elif meso.ScanDone() & key:
   pipe, src, dest = 'meso', meso.ScanDone, ScanDone.Meso
else:
 PipelineException
```

dimitri-yatsenko: The for loop has more advantages.  The src is specified only once whereas in the if-elif-...-else structure, it has to be specified twice and that's error prone.  Same thing about spelling out pipe, src, dest multiple times.  I think the for loop is cleaner, especially if you add more pipelines.

python/pipeline/notify.py
```diff
@@ -28,15 +28,18 @@ class SlackUser(dj.Manual):
...
-            from slacker import Slacker
-            api_key, user = (self * SlackConnection()).fetch1('api_key','slack_user')
-            s = Slacker(api_key, timeout=60)
-
-            if message: # None or ''
-                s.chat.post_message('@' + user, message, as_user=True)
-            if file is not None:
-                s.files.upload(file_=file, channels='@' + user,
-                               title=file_title, initial_comment=file_comment)
+            try:
```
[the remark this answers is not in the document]
dimitri-yatsenko: The reason I had to do it is that I had trouble installing slacker, not sure why.

PR #281, opened 2017-10-24
Parallel MotionCorrection, SummaryImages, Quality and memmap creation (in Segmentation)

ecobost:
Every step runs fine in under 32 GB of RAM (segmentation may take a bit more, under 64 anyway). They should also be faster, in my own computer it was constrained by the network bandwidth (120 MB/s), and I haven't run it in at-compute002/3 yet which has higher bandwidth.
Other minor change: it now ignores exceptions that happen during slack notifications (sometimes it timed out or it had network problems).


python/pipeline/meso.py
```diff
@@ -266,56 +268,73 @@ def _make_tuples(self, key):
...
+            # Correct mini scan
             correct_raster = (RasterCorrection() & key & {'field': field_id + 1}).get_correct_raster()
-            scan_ = correct_raster(scan_)
-            scan_ -= scan_.min()  # make nonnegative for fft
+            mini_scan = correct_raster(mini_scan)
 
             # Create template
-            middle_frame = int(np.floor(scan.num_frames / 2))
-            mini_scan = scan_[:, :, max(middle_frame - 1000, 0): middle_frame + 1000]
-            mini_scan = 2 * np.sqrt(mini_scan + 3 / 8)  # *
```
dimitri-yatsenko: comment Anscombe transform

ecobost: ?, just put anscombe transform, rather than the * and the explanation below?

dimitri-yatsenko: Oh, i see.  I just did not understand the `# *` notation

PR #282, opened 2017-10-25
minor improvements in fuse.py

dimitri-yatsenko:
I changed the version requirement for the datajoint to 0.9.0+ because it will have unions.  0.8.2 is about to be release without the union feature.


python/pipeline/fuse.py
```diff
@@ -64,7 +62,10 @@ class Activity(Resolver, dj.Computed):
     -> Pipe
     """
 
-    key_source = meso.Activity().proj() + reso.Activity().proj()
```
ecobost: For completeness could you also do it in the ScanDone keysource

PR #283, opened 2017-10-27
added matlab classes for +stack schema

jakereimer:


ecobost:
can you delete the old stack schemas: stack.Version should probably be updated; stack.RasterCorrection, stack.MotionCorrection, stack.Alignment, stack.Stitching don't exist anymore

ecobost:
I added this changes in PR #287

PR #289, opened 2017-11-29
Anatomy schema + Movies analysis schema + latest OptImageBar

olakiril:


matlab/schemas/+anatomy/Area.m
```diff
@@ -0,0 +1,13 @@
+%{
+# LUT for Brain Areas
+brain_area      :  enum('V1','P','POR','PM','AM','A','RL','AL','LI','LM')      # area name
+---
+%}
+
+classdef Area < dj.Lookup
```
ecobost: there is also a similar table called experiment.BrainArea, are both necessary?

matlab/schemas/+anatomy/Layer.m
```diff
@@ -0,0 +1,12 @@
+%{
+# 
+layer                       : char(12)                      # short name for cortical area
```
ecobost: same here with experiment.Layer

matlab/schemas/+map/OptImageBar.m
```diff
@@ -20,69 +20,61 @@
...
+                        Data = squeeze(mean(reader(:,:,1,:,:),4));
+                        Data = permute(Data,[3 1 2]);
+                        [nslices, data_fs] = fetch1(preprocess.PrepareGalvo & key,'nslices','fps');
+                    end
                     
                     % calculate frame times
-                    frame_times = fetch1(preprocess.Sync & key,'frame_times');
+                    frame_times = fetch1(stimulus.Sync & key,'frame_times');
                     frame_times = frame_times(1:nslices:end);
+                    frame_times = frame_times(1:size(Data,1));
```
ecobost: Not sure why the double downsampling of frameTimes here, I thought downsampling using nSlices (proper slices no depths) should be enough

matlab/schemas/+map/OptImageBar.m
```diff
@@ -20,69 +20,61 @@
...
-                    reader = preprocess.getGalvoReader(key);
-                    Data = squeeze(mean(reader(:,:,1,:,:),4));
-                    Data = permute(Data,[3 1 2]);
-                    [nslices, data_fs] = fetch1(preprocess.PrepareGalvo & key,'nslices','fps');
+                    if strcmp(setup,'2P4') % mesoscope
+                        path = getLocalPath(fullfile(path, sprintf('%s*.tif', name)));
+                        reader = ne7.scanreader.readscan(path,'int16',1);
+                        Data = permute(squeeze(mean(reader(:,:,:,:,:))),[3 1 2]);
+                        data_fs = reader.fps;
+                        nslices = length(reader.fieldDepths);
```
ecobost: this will give back number of fields: fieldDepths is a list with corresponding depths per field. reader.nScanningDepths should give you number of slices (or reader.nFields if you actually do want the number of fields).

matlab/schemas/+map/OptImageBar.m
olakiril: If the scan did not complete, not all fields have the same number and the frame_times is longer

matlab/schemas/+anatomy/Area.m
olakiril: Its better if we move them to anatomy schema.

matlab/schemas/+anatomy/Layer.m
olakiril: this as well

matlab/schemas/+anatomy/Area.m
```diff
@@ -0,0 +1,13 @@
...
+	methods
+		function self = Area(varargin)
```
dimitri-yatsenko: Why do we need a restriction in the constructor?

matlab/schemas/+anatomy/AreaMask.m
```diff
@@ -0,0 +1,65 @@
...
+            
+            [masks,areas] = fetchn(obj,'mask','area');
+            vessels = fetchn(map.OptImageBar & obj,'vessels');
+            
+            area_map = zeros(size(masks{1}));
+            for imasks = 1:length(masks)
+                area_map(masks{imasks}) = imasks;
+            end
+            
+            im(:,:,1) = normalize(area_map);
```
dimitri-yatsenko: `cat(3, ...)` may be more explicit here.

matlab/schemas/+anatomy/AreaMembership.m
```diff
@@ -0,0 +1,67 @@
...
+                % fetch cell coordinates
+                if strcmp(setup,'2P4')
+                    [px_width, px_height, keys] = ...
+                        fetchn((meso.ScanSetUnitInfo & (fuse.ScanSetUnit & field_key)) * ...
+                        proj(meso.ScanInfoField & fuse.ScanSet &  (fuse.ScanSetUnit & field_key),'px_height','px_width'),...
+                        'px_x','px_y');
+                else
+                    [px_width, px_height, keys] = ...
+                        fetchn((reso.ScanSetUnitInfo & (fuse.ScanSetUnit & field_key)) * ...
+                        proj(reso.ScanInfo & fuse.ScanSet &  (fuse.ScanSetUnit & field_key),'px_height','px_width'),...
```
dimitri-yatsenko: `& fuse.ScanSet` is unnecessary here.  And so is `& (fuse.ScanSetUnit & field_key)` -- it's already included in the left side of the join.

matlab/schemas/+map/OptImageBar.m
```diff
@@ -239,34 +239,92 @@ function plotTight(obj,varargin)
             if isempty(keys); disp('Nothing found!'); return; end
             
             for ikey = 1:length(keys)
```
dimitri-yatsenko: Try `for key = keys'` instead of indexing with `ikey`

matlab/schemas/+map/OptImageBar.m
```diff
@@ -239,34 +239,92 @@ function plotTight(obj,varargin)
...
+                im(1:size(h,1),1:size(h,2),2) = zeros(size(v));
                 im(1:size(h,1),1:size(h,2),3) = v;
-
+                
                 im(size(h,1)+1:end,1:size(h,2),1) = zeros(size(v));
                 im(size(h,1)+1:end,1:size(h,2),2) = zeros(size(v));
                 im(size(h,1)+1:end,1:size(h,2),3) = v;
-
+                
                 im(1:size(h,1),size(h,2)+1:end,1) = h;
```
dimitri-yatsenko: You can do this in one line
```matlab
im(1:size(h,1), size(h,2)+1,:) = cat(3, h, s, v)
```

matlab/schemas/+anatomy/Area.m
olakiril: When I wanted to use Area('brain_area="V1"')

matlab/schemas/+anatomy/Area.m
```diff
@@ -0,0 +1,13 @@
+%{
+# LUT for Brain Areas
+brain_area      :  enum('V1','P','POR','PM','AM','A','RL','AL','LI','LM')      # area name
```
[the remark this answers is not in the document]
olakiril: The field was varchar in the database. I am not sure why this was not updated. Fixed now.

matlab/schemas/+anatomy/AreaMembership.m
olakiril: Fixed

matlab/schemas/+anatomy/AreaMask.m
olakiril: Fixed

matlab/schemas/+map/OptImageBar.m
olakiril: I need key

matlab/schemas/+map/OptImageBar.m
olakiril: fixed

matlab/schemas/+anatomy/Area.m
dimitri-yatsenko:
```matlab
Area & 'brain_area="V1"
```

matlab/schemas/+map/OptImageBar.m
dimitri-yatsenko: Not really but this is not a big deal

matlab/schemas/+anatomy/AreaMask.m
```diff
@@ -0,0 +1,63 @@
...
+-> anatomy.Area
+field                    : tinyint               # slice id
+---
+mask                     : mediumblob            # mask of area
+%}
+
+classdef AreaMask < dj.Imported
+    methods(Access=protected)
+        function makeTuples(obj,key) %create clips
+            insert( obj, key );
```
dimitri-yatsenko: `field` is not entered (`slice_id`)   and the comment seems to not match.  This `makeTuples` seems wrong.  If it's not implemented, I would just put `error 'not implemented' rather than suggest that it does something.

olakiril: It is fine.

PR #290, opened 2017-12-01
stack Pipeline

ecobost:
Stacks are raster corrected,, motion corrected, stitched (if they overlap) and aligned (from slice to slice)


ecobost:
I'll close this pull request until I add some changes that we discussed yesterday

PR #295, opened 2017-12-11
Changes in notify and pupil tracking

fabiansinz:


python/pipeline/meso.py
```diff
@@ -1489,6 +1489,14 @@ def get_all_centroids(self, centroid_type='um'):
             centroids = np.stack([xs, ys], axis=1)
         return centroids
 
+@schema
+class ManualDepth(dj.Manual):
```
ecobost: did we decide on anything about these depths? this should be done automatically

PR #296, opened 2017-12-12
No smoothing of motion shifts

ecobost:
and some name attribute changes in MotionCorrection table

### DIFF
--- a/python/pipeline/utils/galvo_corrections.py
+++ b/python/pipeline/utils/galvo_corrections.py
@@ -73,7 +73,8 @@ def compute_motion_shifts(scan, template, in_place=True, num_threads=8,
     :param float outlier_threshold: Threshold for outlier detection (number of stddevs).
 
     :returns: (y_shifts, x_shifts) Two arrays (num_frames) with the y, x motion shifts
-    :returns: (outliers) A boolean array (num_frames) with True for outlier frames.
+    :returns: (outliers) A boolean array (num_frames) with True for outlier frames. Or
+        None if fix_outliers is False.
 
     ..note:: Based in imreg_dft.translation().
     """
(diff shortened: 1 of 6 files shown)


python/pipeline/utils/galvo_corrections.py
```diff
@@ -57,8 +57,7 @@ def compute_raster_phase(image, temporal_fill_fraction):
 
 
 def compute_motion_shifts(scan, template, in_place=True, num_threads=8,
-                          fix_outliers=True, outlier_threshold=0.05, smooth_shifts=True,
-                          smoothing_window_size=5):
+                          fix_outliers=True, outlier_threshold=5):
```
jiakunf: Where is `fix_outliers` param used?

ecobost: yes, it should be used. Fixed below

python/pipeline/utils/performance.py
```diff
@@ -406,12 +406,18 @@ def parallel_motion_stack(chunks, results, raster_phase, fill_fraction,
         if abs(raster_phase) > 1e-7:
             field = galvo_corrections.correct_raster(field, raster_phase, fill_fraction)
 
+        # Compute initial template by averaging 10 frames that correlate highly with middle one
+        num_frames = field.shape[-1]
+        frames = field.reshape((-1, num_frames)) # num_pixels x num_frames
+        residuals = frames - frames.mean(axis=0)
+        frames_std = frames.std(axis=0)
+        covs = np.mean(residuals.T * residuals[:, int(num_frames / 2)], axis=-1)
```
jiakunf: Should it be `@` operator instead of `*`? Otherwise the size of covariance matrix will be num_frames * num_pixels, and so is the correlation matrix. I think they should both be vectors.

ecobost: I'm calculating here only one row of the correlation matrix so covs and corr are a single [num_frames] vector: covariance of each i with the middle frame and correlations of each i with the middle frame. The way to get the entire [num_frames x num_frames] correlation matrix will be something like `np.dot(residuals.T, residuals)/(frames.shape[0] * np.outer(frames_std, frames_std))` which looks nicer but is slightly more work. Here residuals.T is [num_frames x num_pixes] and residuals[:, int(num_frames / 2) is [num_pixels], the * product will return a num_frames x num_pixels and I take the mean over all pixels to get the covariances per frame it's a bit dirty but I think it works

PR #299, opened 2017-12-22
small improvement in stack motion correction

ecobost:
Generate the initial template for motion correction by computing the correlation between frames, selecting the frame that has the highest mean correlation with all others frames (to get a frame with good signal in it) and averaging the 10 frames that correlate highly with the selected frame. Before I just selected the middle one but that frame could be entirely dark or too noisy

### DIFF
--- a/python/pipeline/utils/performance.py
+++ b/python/pipeline/utils/performance.py
@@ -407,14 +407,17 @@ def parallel_motion_stack(chunks, results, raster_phase, fill_fraction, skip_row
         # Apply anscombe transform
         field = 2 * np.sqrt(field - field.min() + 3 / 8)
 
-        # Compute initial template by averaging 10 frames that correlate highly with middle one
-        num_frames = field.shape[-1]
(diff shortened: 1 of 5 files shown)


python/pipeline/utils/performance.py
```diff
@@ -407,14 +407,17 @@ def parallel_motion_stack(chunks, results, raster_phase, fill_fraction, skip_row
         # Apply anscombe transform
         field = 2 * np.sqrt(field - field.min() + 3 / 8)
 
-        # Compute initial template by averaging 10 frames that correlate highly with middle one
-        num_frames = field.shape[-1]
-        frames = np.reshape(field[skip_rows:-skip_rows, skip_cols:-skip_cols], (-1, num_frames)) # num_pixels x num_frames
+        # Compute correlation matrix between frames (ignoring edges)
```
dimitri-yatsenko: These reshapings are not necessary when adopting np.tensordot instead of np.dot.

PR #303, opened 2018-01-23
minor

dimitri-yatsenko:
minor fix for datajoint 0.10.0+ compatibility

Starting in datajoint 0.10.0,  restriction of a relation with another relation cannot be done on attributes that are not part of the primary key in either of the operands.


dimitri-yatsenko:
@ecobost, if there was any code joining on non-primary attributes, it was probably wrong and I don't think we have had many such cases.  This particular example is made more efficient by excluding the non-primary attributes from the restriction.

PR #314, opened 2018-02-19
matlab classes

jakereimer:


matlab/schemas/+stack/StackSet.m
```diff
@@ -0,0 +1,18 @@
+%{
```
ecobost: I'm still testing this one and it's probably gonna change

matlab/schemas/+stack/Corrections.m
```diff
@@ -0,0 +1,18 @@
+%{
```
ecobost: All these starting with Corrections (Corrections, CorrectionsRaster, CorrectionsMotion, CorrectionsStitched, CorrectionsROICoordinates) are from an older version, they are not in the code or the database anymore

PR #318, opened 2018-03-02
Set of small improvements

ecobost:
- Fix bug in size of FieldInStack during registration.
- Add StackSet to create unique ids per stack
- Reorder populate script
- Allow stitching with a min of 10 pixels / before min was half width of smaller field

### DIFF
--- a/matlab/scripts/populate_sync.m
+++ b/matlab/scripts/populate_sync.m
@@ -1,7 +1,4 @@
 while 1
-	try
-        parpopulate(stimulus.Sync, experiment.AutoProcessing)
-        pause(1000);
-	catch ME
-    end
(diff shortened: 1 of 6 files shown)


matlab/scripts/populate_sync.m
```diff
@@ -0,0 +1,7 @@
+while 1
+	try
+        parpopulate(stimulus.Sync, experiment.AutoProcessing)
+        pause(1000);
```
dimitri-yatsenko: rather than `try ... catch`, get a return argument from `parpopulte`:
```matlab
k = parpopulate(stimulus.Sync, experiment.AutoProcessing)
```
This has the same effect as `suppress_errors` in datajoint-python.

ecobost: oh, ok, thanks

PR #319, opened 2018-03-06
add table for posture camera preview

jiakunf:
New table stores posture camera total number of frames and 16 preview frames.

### DIFF
--- a/python/pipeline/posture.py
+++ b/python/pipeline/posture.py
@@ -1,0 +1,106 @@
+from scipy.misc import imresize
+import datajoint as dj
+from datajoint.jobs import key_hash
+
+from . import experiment, notify
+from .exceptions import PipelineException
+
+from warnings import warn
+import cv2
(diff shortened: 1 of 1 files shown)


python/pipeline/posture.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def _make_tuples(self, key):
+        key['posture_time'], key['preview_frames'], key['total_frames'] = self.grab_timestamps_and_frames(key)
+
+        self.insert1(key)
+        del key['posture_time']
+        frames = key.pop('preview_frames')
+        self.notify(key, frames)
+
+    def notify(self, key, frames):
```
ecobost: Can you add @notify.ignore_exceptions as a decorator in this function (like https://github.com/ecobost/pipeline/blob/master/python/pipeline/reso.py#L130); it's just so a failuer during notification doesn't roll back the insert.

PR #336, opened 2018-08-09
updated dockerfile (has jupyter), some bugfixes, delete previous tutorial

ecobost:


Dockerfile.base
```diff
@@ -1,6 +1,7 @@
-FROM ubuntu:17.10
+FROM ubuntu
```
eywalker: why not specific version of Ubuntu? Not sure if pulling from the latest always is a good idea.

ecobost: I don't mind, we need at least 17.10 so I'll base it on 18.04 ubuntu

PR #345, opened 2018-12-14
Stack Registration and Segmentation

ecobost:


python/pipeline/meso.py
```diff
@@ -335,46 +333,34 @@ def _make_tuples(self, key):
         scan_filename = (experiment.Scan() & key).local_filenames_as_wildcard
         scan = scanreader.read_scan(scan_filename, dtype=np.float32)
 
-        for field_id in range(scan.num_fields):
```
DonnieKim411: Why are we not applying `RasterCorrection` on all fields?

python/pipeline/meso.py
```diff
@@ -416,86 +401,85 @@ def _make_tuples(self, key):
         scan_filename = (experiment.Scan() & key).local_filenames_as_wildcard
         scan = scanreader.read_scan(scan_filename)
 
-        for field_id in range(scan.num_fields):
```
DonnieKim411: same deal as above. Why not all the fields?

python/pipeline/meso.py
```diff
@@ -624,56 +607,59 @@ def _make_tuples(self, key):
...
```
DonnieKim411: Same as above. for loop for `scan.num_fields` is missing

python/pipeline/stack.py
```diff
@@ -408,7 +407,7 @@ def _make_tuples(self, key):
             f = performance.parallel_motion_stack # function to map
             raster_phase = (RasterCorrection() & key).fetch1('raster_phase')
             fill_fraction = (StackInfo() & key).fetch1('fill_fraction')
-            max_y_shift, max_x_shift = 20 / (StackInfo.ROI() & key).microns_per_pixel
+            max_y_shift, max_x_shift = 20 / (StackInfo.ROI() & key).microns_per_pixel[1:]
```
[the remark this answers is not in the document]
ecobost: it's just a threshold I choose for max shift (by asking Jake and I only use the second axes on, because the first axes is depth/z and we only have shifts in y and x

python/pipeline/stack.py
```diff
@@ -512,7 +511,6 @@ class ROICoordinates(dj.Part):
         -> Stitching.Volume             # volume to which this ROI belongs
         stitch_xs        : blob         # (px) center of each slice in the volume-wise coordinate system
         stitch_ys        : blob         # (px) center of each slice in the volume-wise coordinate system
-        stitch_z         : float        # (um) initial depth in the motor coordinate system
```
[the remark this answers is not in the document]
ecobost: Because I'm not actually changing the z when I do the stitching. I only stitch in y and x.  But two field of views that were supposedly recorded at the same depth are assumed to be at the same depth. This is ok for mesoscope scans (the main and only use for stitching till now): two overlapping field-of-views may be a bit off in x, y but they are pretty much the same in z

python/pipeline/meso.py
ecobost: Before corrections (raster and motion) were populated once per scan (and thus we needed to run a for loop over all fields, now I ran the populate once per field;  it makes it faster because all fields can be motion corrected in parallel (current bottleneck)

python/pipeline/meso.py
ecobost: same answer

python/pipeline/meso.py
ecobost: same answer. Now the populate is ran once per field and not once per scan so no need to loop over all fields

PR #347, opened 2019-03-01
legacy import corrections and manual classification gui (unfinished)

paulgfahey:


python/pipeline/utils/mask_classification.py
```diff
@@ -60,4 +61,219 @@ def on_button(event):
...
+        fig.canvas.mpl_connect('key_press_event', on_button)
+
+        plt.show()
+
+    sns.reset_orig()
+
+    return mask_types
+
+
+def detect_peaks(x, mph=None, mpd=1, threshold=0, edge='rising',
```
ecobost: is this the same as the one in stimline. Maybe add it in pipeline/utils/signal.py and import it in stimline (as `from pipeline.utils import signal; signal.detect_peaks` rather than have it in both places.

python/pipeline/meso.py
```diff
@@ -1040,7 +1040,7 @@ def update_img(i):
             print('If this takes too long, stop it and call again with dpi <', dpi, '(default)')
             video.save(filename, dpi=dpi)
 
-            return fig
+            return fig,video
```
ecobost: either document that it willl also return video and add the return for reso too to keep it consistent or only return figure. I think video dies when you leave the function and that's why I didn't return it

python/pipeline/utils/mask_classification.py
ecobost: Also there is a find_peaks in signal.py that may be enough for whatever you are doing (or maybe not)

paulgfahey: will delete from stimline, also is still work in progress

python/pipeline/meso.py
paulgfahey: video removed from return

PR #357, opened 2019-04-11
Added surgery tables to experiment.py

Neuro-CS:
Added several new Manual and Lookup tables for the inputting/editing of surgery entries via pipecontrol.


python/pipeline/experiment.py
```diff
@@ -333,6 +333,73 @@ def plot_calibration_curve(self):
         return fig
 
 
+@schema
+class SurgeryTypes(dj.Lookup):
```
dimitri-yatsenko: Just as a convention in the lab, the class name reflects the type of entries in each row of the table. So in this case it should be `SurgeryType`, etc in the singular form.  The same goes for other tables.

python/pipeline/experiment.py
```diff
@@ -333,6 +333,73 @@ def plot_calibration_curve(self):
...
+@schema
+class SurgeryStatus(dj.Manual):
+    definition = """ # Updates to surgeries performed on mice
+    -> Surgery
+    timestamp                           : timestamp              # Timestamp of entry
+    ---
+    euthanized = "False"                : enum("True","False")
+    day_one = "False"                   : enum("True","False")   # First day checkup performed
+    day_two = "False"                   : enum("True","False")   # Second day checkup performed
+    day_three = "False"                 : enum("True","False")   # Third day checkup performed
```
dimitri-yatsenko: If there are multiple checkups, then they belong in a different table, e.g. `SurgeryCheckup`

Neuro-CS: There are three checkups that need to be performed. Every time the user modifies the surgery status, a newer entry is added with the values specified. All processing takes place on the newest entry. If checkups should become a new table, how would you suggest the table be structured?

python/pipeline/experiment.py
Neuro-CS: Got it! I've updated the class names.

user
```diff
@@ -0,0 +1,68 @@
+<!DOCTYPE html>
```
ecobost: Can you drop this file?

python/pipeline/experiment.py
```diff
@@ -333,6 +333,73 @@ def plot_calibration_curve(self):
...
+    """
+
+
+@schema
+class SurgeryStatus(dj.Manual):
+    definition = """ # Updates to surgeries performed on mice
+    -> Surgery
+    timestamp                           : timestamp              # Timestamp of entry
+    ---
+    euthanized = "False"                : enum("True","False")
```
ecobost: you can also use type boolean and datajoint will internally convert it to tinyint (0/1 flag) and retrieve it as boolean (not 100% sure about the retrieval but it's probably more standard than an enum(False, True))

python/pipeline/experiment.py
```diff
@@ -333,6 +333,73 @@ def plot_calibration_curve(self):
...
+    contents = [
+        ['Survival'],
+        ['Non-Survival'],
+    ]
+
+
+@schema
+class SurgeryQuality(dj.Lookup):
+    definition = """ # Surgery qualities used in experiment.Surgery
+    surgery_quality              : varchar(32)         # Possible qualities of surgeries performed on mice
```
ecobost: Could we better have something like a 1-5 or 1-100 ranking (that way we don't need this table, it will just be an int), it will still be qualitative but maybe finer grained.

python/pipeline/experiment.py
ecobost: I still don't understand quite what you mean there. What I get is that they will sometimes checkup on the mouse surgery and optionally euthanize it. I would just have something like 
```python
@schema
class SurgeryStatus(dj.Manual):
    definition = """
    -> Surgery
    timestamp    :datetime
    ----
    euthanized    : boolean                           # was the mouse euthanized during this checkup
    """
```
that is the db friendly way.  You just insert a new entry for each checkup
If they wanna know how many checkups a mice has had, they'll have to check `len(SurgeryStatus & surgery_key)` or you can offer the `num_checkups` attribute via a property:

```python
@schema
class SurgeryStatus(dj.Computed ):
    definition = """.... """

    @property
    def num_checkups(self):
        return len(self)
# And people will call it like
(experiment.SurgeryStatus & key).num_checkups # returns a number
```
Alternatively, in SQL there's the concept of a View which is a virtual table with data computed on-the-fly from actual real tables (e.g.,  num_checkups computed from the number of rows in SurgeryStatus), we don't have that in python-datajoint but you could do something like:
```python
@schema
class SurgeryStatus(dj.Computed ):
definition = """....
"""
    def num_checkups():
        """ Number of checkups for each surgery in SurgeryStatus"""
        return SurgeryStatus.aggr(Surgery, num_checkups="COUNT(*)"
# And people will call it like
experiment.SurgeryStatus.num_checkups() # returns a table that has all primary keys of Surgery plus the attribute num_checkups
```

Neuro-CS: Those three entries for checkups are there because everyone is required to check the mouse the next three days after a surgery. The reason I don't go off timestamps is because someone could check the mouse but forget to make a new entry that day.

user
Neuro-CS: Whoops. Deleted it.

python/pipeline/experiment.py
```diff
@@ -359,28 +359,17 @@ class SurgeryOutcome(dj.Lookup):
...
-        ['Okay'],
-        ['Bad']
-    ]
-
-
 @schema
 class Surgery(dj.Manual):
     definition = """ # List of surgeries performed on mice
     -> mice.Mice
     date                         : date                   # Date surgery was performed
```
ecobost: Maybe instead of having the timestamp of the surgery as primary key, you can use an int (surgery_id) that just increases for every surgery. Not sure whether it's recommended to have times in primary keys (quick Google search is ambivalent). It can still be a secondary attribute.
It also makes restrictions nicer `Surgery & {'animal_id': 123, 'surgery_id': 0}` rather than `Surgery & {'animal_id': 123, 'surgery_id': '2019-08-03'}` which requires knowing date/time/timestamp formatting (for instance, I'm not sure the restriction I just wrote will work or whether 08 is the month or the date)

ecobost: Also, not sure about your use of date and time as diff attributes, I would just use a timestamp as we do pretty much everywhere we want to record time.

Neuro-CS: Restrictions would be difficult to perform without knowing the exact surgery_id ahead of time, so I'd end up using the secondary 'date' key for the restrictions in most use cases. I can add a description to the date key saying "YYYY-MM-DD format" to help clear up that confusion.

PR #361, opened 2019-04-19
Making deeplabcut pupil tracking as a permanent method in pipeline

DonnieKim411:


Dockerfile.base
```diff
@@ -93,6 +65,22 @@ RUN apt-get install autoconf automake libtool && \
     rm -r cmt && \
     pip3 install git+https://github.com/cajal/c2s.git # stm (spike-triggered mixture model)
 
+###############################################################################
+# install Deeplabcut and its dependencies
+RUN pip3 install git+https://github.com/cajal/DeepLabCut.git
+#RUN pip3 install https://extras.wxpython.org/wxPython4/extras/linux/gtk3/ubuntu-16.04/wxPython-4.0.3-cp36-cp36m-linux_x86_64.whl
+
+# Install dependencies of wxPython
```
ecobost: you can delete this comments

Dockerfile.base
```diff
@@ -15,37 +15,9 @@ RUN apt-get update && \
     apt-get install -y python3-dev python3-pip python3-numpy python3-scipy \
         python3-matplotlib
 
-###############################################################################
-# Install OpenCV 3.4.3 with CUDA support (take a seat, this takes hours)
```
ecobost: where is opencv, now? is it installed by tensorflow?

docker-compose.yml
```diff
@@ -1,14 +1,15 @@
-version: '2'
+version: '2.3'
```
ecobost: you should leave the original docker-compose here

python/pipeline/.ipynb_checkpoints/exceptions-checkpoint.py
```diff
@@ -0,0 +1,6 @@
+class PipelineException(Exception):
```
ecobost: Can you delete this .ipynb_checkpoints from the PR?

python/pipeline/reso.py
```diff
@@ -1051,12 +1051,12 @@ class CNMFBackground(dj.Part):
...
 
-    def _make_tuples(self, key):
+    def make(self, key):
         # Create masks
         if key['segmentation_method'] == 1:  # manual
-            Segmentation.Manual()._make_tuples(key)
+            Segmentation.Manual().make(key)
         elif key['segmentation_method'] in [2, 6]:  # nmf and nmf-patches
-            Segmentation.CNMF()._make_tuples(key)
+            Segmentation.CNMF().make(key)
```
ecobost: you did not change the stuff inside CNMF here (and the allow _direct insert in Fluorescence)

python/pipeline/meso.py
```diff
@@ -918,9 +918,10 @@ def _make_tuples(self, key):
             dj.conn()
 
             ## Insert in CNMF, Segmentation and Fluorescence
-            Segmentation().insert1(key)
-            Segmentation.CNMF().insert1({**key, 'params': json.dumps(kwargs)})
-            Fluorescence().insert1(key)  # we also insert traces
+            # Segmentation().insert1(key)
```
ecobost: you can delete this comments

python/pipeline/meso.py
```diff
@@ -918,9 +918,10 @@ def _make_tuples(self, key):
...
+            # Segmentation.CNMF().insert1({**key, 'params': json.dumps(kwargs)})
+            self.insert1({**key, 'params': json.dumps(kwargs)})
```
ecobost: I would rather have Segmentation.insert1 here (even with allow_direct_insert) so it parallels the Manual insertions (i.e, CNMF). Either that or down there in the main, have self.insert1(key) for Manual too.

python/scripts/populate-minion.py
```diff
@@ -35,7 +35,7 @@
             pipe.MotionCorrection().populate(next_scans, reserve_jobs=True, suppress_errors=True)
             pipe.SummaryImages().populate(next_scans, reserve_jobs=True, suppress_errors=True)
             pipe.Segmentation().populate(next_scans, reserve_jobs=True, suppress_errors=True)
-            pipe.Fluorescence().populate(next_scans, reserve_jobs=True, suppress_errors=True)
+            # pipe.Fluorescence().populate(next_scans, reserve_jobs=True, suppress_errors=True)
```
ecobost: why is this commented out?

python/scripts/test_script.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/local/bin/python3
```
ecobost: don;t need this

python/tests/.ipynb_checkpoints/test_preprocess-checkpoint.py
```diff
@@ -0,0 +1,126 @@
+""" Test suite for pre processing routines."""
```
ecobost: make sure this .ipynb is also gone

python/pipeline/utils/DLC_tools.py
```diff
@@ -0,0 +1,1033 @@
...
+
+        # dpi=self.dpi, fps=self.clip.FPS
+        save_video_path = self.video_path.split('.')[0] + '_fitted_' + \
+                str(start) + '_' + str(end) + '_labeled.avi'
+        ani.save(save_video_path, writer=writer, dpi=self.dpi)
+
+        return ani
+
+
+# # We just subclass Rectangle so that it can be called with an Axes
```
ecobost: do you need all of these comments?, they are in your git history so they are there if you wanna come back to them but may not need them in the main repo

python/pipeline/utils/DLC_tools.py
```diff
@@ -0,0 +1,1033 @@
...
+
+    if result is None:
+        raise AssertionError()
+    return result
+
+class PlotBodyparts():
+
+    def __init__(self, config, bodyparts='all'):
+        """
+        Input:
```
ecobost: the standard docstring uses Arguments: and Returns: rather than Input and Output (like here https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). You don't need to change it, just for future reference

python/pipeline/pupil_new.py
```diff
@@ -0,0 +1,493 @@
+import os
```
ecobost: do we need pupil_new?

python/pipeline/pupil.py
```diff
@@ -540,4 +575,463 @@ def make(self, key):
 
         self.insert1(key)
         for ckey in tqdm(contours):
-            self.Ellipse().insert1(ckey, ignore_extra_fields=True)
\ No newline at end of file
+            self.Ellipse().insert1(ckey, ignore_extra_fields=True)
+
+# If config.yaml ever updated, make sure you store the file name differently so that it becomes unique
```
ecobost: I didn't really checked the code below but I trust you did :)

python/pipeline/pupil.py
```diff
@@ -540,4 +575,463 @@ def make(self, key):
...
+        compressed_cropped_video.insert1(dict(key, cropped_x0=cropped_coords['cropped_x0'],
+                                              cropped_x1=cropped_coords['cropped_x1'],
+                                              cropped_y0=cropped_coords['cropped_y0'],
+                                              cropped_y1=cropped_coords['cropped_y1'],
+                                              added_pixels=pixel_num,
+                                              video_path=compressed_cropped_video_path))
+
+
+@schema
+class FittedContourDeeplabcut(dj.Computed):
```
ecobost: Sure we want this to be a different table rather than a different method on whatever table we already used? Did you check with somebody about that? They'll both require changes in downstream code but not sure which is the best way to go about it

Dockerfile.base
DonnieKim411: deleted

Dockerfile.base
DonnieKim411: It is now installed by deeplabcut as a part of the package

docker-compose.yml
DonnieKim411: If I am not mistaken, if you wanna use `nvidia` as `runtime`, then the version must be 2.3, no?

python/pipeline/.ipynb_checkpoints/exceptions-checkpoint.py
DonnieKim411: Deleted. Somehow some of the `.ipynb_checkpoints` are not completed ignored even after putting into `.gitignore`

python/pipeline/reso.py
DonnieKim411: Updated

python/pipeline/meso.py
DonnieKim411: Deleted

python/scripts/populate-minion.py
DonnieKim411: I commented this out because we are populating `Flourescence()` when we populate  `pipe.Segmentation()`. So we are calling `Flouresence` for no reason

python/scripts/test_script.py
DonnieKim411: Deleted

python/tests/.ipynb_checkpoints/test_preprocess-checkpoint.py
DonnieKim411: Deleted

python/pipeline/utils/DLC_tools.py
DonnieKim411: I deleted all these per request

python/pipeline/utils/DLC_tools.py
DonnieKim411: Got it.

python/pipeline/pupil_new.py
DonnieKim411: deleted

python/pipeline/pupil.py
DonnieKim411: OK. In my mind, things are at least logically structured...

docker-compose.yml
ecobost: we can change the versions but all other attributes should be the same as before (or at least not point to your folders)

python/scripts/populate-minion.py
ecobost: Fluorescence will still need to be computed for Manual masks

python/pipeline/pupil.py
```diff
@@ -540,4 +575,998 @@ def make(self, key):
...
+    definition = """
+    # Minimal info needed to load deeplabcut model
+    config_path         : varchar(255)          # path to deeplabcut config
+    ---
+    shuffle             : smallint unsigned     # shuffle number used for the trained dlc model. Needed for dlc.analyze_videos
+    trainingsetindex    : smallint unsigned     # trainingset index used for the trained dlc. model. Needed for dlc.analyze_videos
+    """
+
+@schema
+class TrackingMethod(dj.Manual):
```
ecobost: for list of stuff that won't change often we use Lookup tables. We have all other methods in pipeline/shared.py, you should probably put this one there and follow the convention from there:
``` python
@schema
class TrackingMethod(dj.Lookup):
    definition = """
    tracking_method : tinyint                   # method used for tracking
    ---
    name                : varchar(16)               # short name to identify the tracking method
    details             : varchar(255)              # more details
    language            : enum('matlab', 'python')  # implementation language
    """

    contents = [
        [1, 'manual', 'sjsksksl', 'python'],
        [2, 'deeplabcut', 'ssllslsl', 'python'],
    ]
```

python/pipeline/pupil.py
```diff
@@ -540,4 +575,998 @@ def make(self, key):
...
+    tracking_method     : tinyint unsigned # tracking_method 0:manual 1: deeplabcut
+    ---
+    """
+
+    def fill(self, key, tracking_method):
+        self.insert1(dict(key, tracking_method=tracking_method))
+
+@schema
+class Tracking(dj.Computed):
+    definition="""
```
ecobost: Eye should be here:
```python
class Tracking(dj.Computed)
 -> Eye
->TrackingMethod
---
tracking_ts
```
And then people can call it as Trakcing.populate({'animal_id': slslls, 'session': slslls, 'scan_idx': sksks, 'tracking_method': slslsl}). And when we add it to the populate script, it will just be `pupil.Tracking(next_scans, {'tracking_method': 1}) #assuming 1 is deeplabcut`

python/pipeline/pupil.py
```diff
@@ -540,4 +575,998 @@ def make(self, key):
...
+    -> TrackingMethod
+    ---
+    tracking_ts=CURRENT_TIMESTAMP   : timestamp  # automatic
+    """
+
+    class ManualTracking(dj.Part):
+        definition="""
+        -> master
+        frame_id                    : int                   # frame id with matlab based 1 indexing
```
ecobost: minor: no need to say matlab based (it's kind of confusing), just say 1-based

python/pipeline/pupil.py
```diff
@@ -540,4 +575,998 @@ def make(self, key):
...
+        ---
+        contour=NULL                : longblob              # eye contour relative to ROI
+        """
+        def make(self, key):
+            print("""
```
ecobost: what I was thinking is to move the make of ManuallyTrackedContours here and put the sign about deprecation in ManuallyTrackedContours, that way all tracking code is in the same place and if you wanna track with deeplabcut or manually you would use the same table (just with a different tracking method). So move all code from ManuallyTrackedContours here, and if somebody tries to populate that table just tell them to populate this with tracking_method: 0 (assuming 0 is manual method); there's only three people (jiakun, taliah and paul) that manually track stuff (and lately it's been only Taliah) so it won't be a problem to ask em to do it this way now. We should also have this function but call it something like migrate (so whatever was in the old ManuallyTracked could be migrated this function)

ecobost: You also need to standardize the saving of results between the two methods. Anything that depends on Tracking shouldn't care how stuff is tracked, it should be able to just read whatever is tracked otherwise there is no use to have them in the same table (for instance we have methods to do diff segmentations or deconvolutions or registrations in the same table but they all save the same results in a shared table, that could be a part table; that way from there on people don't need to care how stuff was segmented or deconvolved, etc.). I was assuming that at some point you transform manual data or deeplabcut data into some shared, standard format that you then fit the circle and ellipse to, this may not be true (should it?), if it isn't I don't see why we are putting it in the same table. It seems like the output of tracking should be the same in both cases, but if it's not they should remain in two distinct tables (because they are not functionally related); the output of fitting will be the same (again, I guess) so maybe there is where we should join stuff

ecobost: I thought we initially were gonna make the joining in the fitted something like:
```python
class ManuallyTrackedContours() # or whatever name it already has
-> Eye
---
manual_tracked _results

class DeepLabCutTrack() 
->Eye
---
deeplabcuttracked_results

class NewFittedContour():
-> Eye
-> shared.TrackingMethod
---
fitted_results
  class ManuallyTrackedFitting(Part):
    definition = """
    -> ManualTrackedContours
    """
    def make():
      # make here will do the fitting bringing data from MAnuallyTrackedContours and inserts the results in NEwFittedContour (
  class DeepLabCutFitted(dj.Part):
    definition = """
    ->DeepLabCutTrack
    """"
    def make():
      # make here again brings stuff from  DeepLabCutTrack and does the fitting
  def make():
    if tracking_method=1:
      self.insert1(key)
      DeepLabcUtftting.make()
    if tracknig_method 2:
      self.insert1(key)
      ManualFitting.make()
```

python/pipeline/data/.ipynb_checkpoints/cnn_model-checkpoint.json
```diff
@@ -0,0 +1,351 @@
+{
```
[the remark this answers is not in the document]
DonnieKim411: I have no idea but i deleted the entire .ipynb_checkpoints directory... idk why this keeps being generated as I have not specified any flag to save the check points. 

Well I just did `diff` between `python/pipeline/data/.ipynb_checkpoints/cnn_model-checkpoint.json` and `python/pipeline/data/cnn_model.json` and they are the exact same copy. So deleting that particular json file as well as its directory doesn't result in any loss. Hence, I am deleting them.

docker-compose.yml
DonnieKim411: Yup, I finally reverted back to the original form... so this shouldnt be a problem anymore

python/scripts/populate-minion.py
DonnieKim411: fixed!

python/tests/.ipynb_checkpoints/test_preprocess-checkpoint.py
DonnieKim411: deleted... but I really gotta find a permanent way to stop generating these checkpoints.

python/scripts/test_script.py
```diff
@@ -0,0 +1,84 @@
+import time
```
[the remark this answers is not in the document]
DonnieKim411: deleted

python/pipeline/pupil.py
```diff
@@ -98,10 +110,11 @@ def _make_tuples(self, key):
                 msg = ('Number of movie frames and timestamps differ: {} frames vs {} '
                        'timestamps'). format(num_video_frames, num_timestamps)
                 raise PipelineException(msg)
-            elif num_timestamps > num_video_frames: # cut timestamps to match video frames
+            elif num_timestamps > num_video_frames:  # cut timestamps to match video frames
                 timestamps_in_secs = timestamps_in_secs[:-1]
-            else: # fill with NaNs
-                timestamps_in_secs = np.array([*timestamps_in_secs, float('nan')])
+            else:  # fill with NaNs
```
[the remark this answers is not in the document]
DonnieKim411: oh yeah. Either VSCode did automatically or I accidentally added a space there... but fully noted

python/pipeline/meso.py
```diff
@@ -1051,12 +1050,13 @@ class CNMFBackground(dj.Part):
...
 
-    def _make_tuples(self, key):
+    def make(self, key):
         # Create masks
         if key['segmentation_method'] == 1:  # manual
-            Segmentation.Manual()._make_tuples(key)
+            Segmentation.Manual().make(key)
         elif key['segmentation_method'] in [2, 6]:  # nmf
-            Segmentation.CNMF()._make_tuples(key)
+            self.insert1(key)
```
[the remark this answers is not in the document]
DonnieKim411: My logical flow on this is the following. First, we have a class `Segmentation` and `CNMF` is another class under `Segmentation`. But according to the Datajoint `insert` flow, `master` table must be inserted first, then the `part` table can be inserted. Hence, line 1058 inserts `master` table first, then in line 921, we insert the `part` table `CNMF`. on line 922, we insert `Fluorescence` table with `allow_direct_insert=True` option. 

I can bring line 1058 outside of all the `if` statements, but it is your call

python/pipeline/reso.py
```diff
@@ -918,10 +918,9 @@ def _make_tuples(self, key):
             dj.conn()
 
             ## Insert in CNMF, Segmentation and Fluorescence
-            Segmentation().insert1(key)
-            Segmentation.CNMF().insert1({**key, 'params': json.dumps(kwargs)})
-            Fluorescence().insert1(key)  # we also insert traces
-
+            self.insert1({**key, 'params': json.dumps(kwargs)})
+            Fluorescence().insert1(key, allow_direct_insert=True)  # we also insert traces
```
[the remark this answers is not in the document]
DonnieKim411: Fluorescence.Trace() is getting inserted on line 939.

Same thing is happening on reso. you can check on https://github.com/DonnieKim411/pipeline/blob/master/python/pipeline/reso.py#L922 and https://github.com/DonnieKim411/pipeline/blob/master/python/pipeline/reso.py#L939

python/pipeline/pupil.py
DonnieKim411: I understand, but this table serves the same purpose as `SegmentationTask` under meso and reso modules. For each key, it simply tells which method to be used either `manual` = 0 and `deeplabcut` = 1. I think it should stay here with the current table definition

python/pipeline/pupil.py
DonnieKim411: Done

python/pipeline/meso.py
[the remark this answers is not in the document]
ecobost: we cannot put line 1058 outside the ifs because when the segmentation is manual it shouldn't be inserted. We can leave it as is for now

python/pipeline/pupil.py
ecobost: we'll leave as is

docker-compose.yml
```diff
@@ -47,8 +47,7 @@ services:
 
   
   bash:
-    image: ninai/pipeline:latest
```
ecobost: this should be stimulus-pipeline:latest and minion-gpu will be pipeline:tf-gpu

PR #364, opened 2019-05-17
Added surface calculation method and table

Neuro-CS:
Added two new tables and paraboloid fit to determine surface of brain in stack.

### DIFF
--- a/python/pipeline/stack.py
+++ b/python/pipeline/stack.py
@@ -6,7 +6,7 @@ import numpy as np
 import scanreader
 from scipy import signal
 from scipy import ndimage
-from scipy.optimize import curve_fit
+from scipy import optimize
 import itertools
 
 from . import experiment, notify, shared, reso, meso
@@ -979,26 +979,11 @@ class PreprocessedStack(dj.Computed):
 
 
 @schema
-class StackSurfaceMethod(dj.Lookup):
-    definition = """ # Methods used to compute surface of the brain
-
-    surface_method_id   : tinyint unsigned   # Unique ID given to each surface calculation method
-    ---
-    method_title        : varchar(32)        # Title of surface calculation method
-    method_description  : varchar(256)       # Details on surface calculation
-    """
-
-    contents = [
-        [1, 'Paraboloid Fit', 'Fit ax^2 + by^2 + cx + dy + f to surface after finding max of sobel']
-    ]
-
-
-@schema
-class StackSurface(dj.Computed):
+class Surface(dj.Computed):
     definition = """ # Calculated surface of the brain
 
     -> PreprocessedStack
-    -> StackSurfaceMethod
+    -> shared.SurfaceMethod
     ---
     surface_im        : longblob          # Matrix of calculated depth for each pixel in stack
     lower_bound_im    : longblob          # Lower bound of 95th percentile confidence interval
@@ -1019,6 +1004,8 @@ class StackSurface(dj.Computed):
         #      - Attempts to replace this with a percentile all fail
         #  - This code drops guessed depths > 95th-percentile and < 5th-percentile to be more robust to outliers
 
+        valid_method_ids = [1]  # Used to check if method is implemented
+
         def surface_eqn(data, a, b, c, d, f):
             x, y = data
             return a * x ** 2 + b * y ** 2 + c * x + d * y + f
@@ -1029,16 +1016,19 @@ class StackSurface(dj.Computed):
         upper_threshold_percent = 0.6  # Surface median intensity should be in the bottom X% of the *range* of medians
         gaussian_blur_size = 5  # Size of gaussian blur applied to slice
         min_points_allowed = 10  # If there are less than X points after filtering, throw an error
-        bounds = (
-        [0, 0, np.NINF, np.NINF, np.NINF], [np.Inf, np.Inf, np.Inf, np.Inf, np.Inf])  # Bounds for paraboloid fit
+        bounds = ([0, 0, np.NINF, np.NINF, np.NINF], [np.Inf, np.Inf, np.Inf, np.Inf, np.Inf])  # Bounds for paraboloid fit
         ss_percent = 0.40  # Percentage of points to subsample for robustness check
         num_iterations = 1000  # Number of iterations to use for robustness check
 
+        if 'surface_method_id' not in key:
+            raise PipelineException('Error: surface_method_id not specified')
+        if int(key['surface_method_id']) not in valid_method_ids:
+            raise PipelineException(f'Error: surface_method_id {key["surface_method_id"]} is not implemented')
+
+
         print('Calculating surface of brain for stack', key)
         full_stack = (stack.PreprocessedStack & key).fetch1('resized')
-        depth = full_stack.shape[0]
-        height = full_stack.shape[1]
-        width = full_stack.shape[2]
+        depth, height, width = full_stack.shape
 
         surface_guess_map = []
         r_xs = np.arange(r, width - width % r, r * 2)[1:-1]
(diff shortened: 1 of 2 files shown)


python/pipeline/stack.py
```diff
@@ -6,6 +6,7 @@
 import scanreader
 from scipy import signal
 from scipy import ndimage
+from scipy.optimize import curve_fit
```
ecobost: I'm gonna be annoying and ask you to import optimize rather than curve_fit directly so `from scipy import optimize` and then use `optimize.curve_fit`, the advantage is that tells me curve_fit is coming from optimize rather than being a function defined by someone (http://google.github.io/styleguide/pyguide.html#22-imports)

python/pipeline/stack.py
```diff
@@ -977,6 +978,129 @@ def make(self, key):
         self.insert1({**key, 'resized': resized, 'lcned': lcned, 'sharpened': sharpened})
 
 
+@schema
```
ecobost: we should put this one in pipeline/shared.py, that's where alll our other method  lookups are defined. Probably also match the convention of the other methods there. They are all kind of like thios:
```python
@schema
class StackSegmMethod(dj.Lookup):
    definition = """ # methods for 3-d stack segmentations
    stacksegm_method         : tinyint
    ---
    name                        : varchar(16)
    details                     : varchar(255)
    language                    : enum('matlab', 'python')  # implementation language
    """
    contents = [
        [1, '3dconv', '3-d convolutional network plus watershed', 'python'],
        [2, '3dconv-ensemble', 'an ensemble of 3-d convolutional networks plus watershed', 'python']
    ]
```
Just to keep stuff consistent

python/pipeline/stack.py
```diff
@@ -977,6 +978,129 @@ def make(self, key):
...
+    method_description  : varchar(256)       # Details on surface calculation
+    """
+
+    contents = [
+        [1, 'Paraboloid Fit', 'Fit ax^2 + by^2 + cx + dy + f to surface after finding max of sobel']
+    ]
+
+
+@schema
+class StackSurface(dj.Computed):
```
ecobost: i would just call it Surface: `stack.Surface` rather than `stack.StackSurface`. Same for StackSurfaceMethod

python/pipeline/stack.py
```diff
@@ -977,6 +978,129 @@ def make(self, key):
...
+    definition = """ # Calculated surface of the brain
+
+    -> PreprocessedStack
+    -> StackSurfaceMethod
+    ---
+    surface_im        : longblob          # Matrix of calculated depth for each pixel in stack
```
ecobost: We don't have a policy of how big something has to be before we put em in an external. This is gonna be a h x w array, right? that's probably ok as a longblob but if you feel like it you can ask in the #datajoint slack channel what they recommend the maximum size per blob (or per row) be

python/pipeline/stack.py
```diff
@@ -977,6 +978,129 @@ def make(self, key):
...
+        #      - Attempts to replace this with a percentile all fail
+        #  - This code drops guessed depths > 95th-percentile and < 5th-percentile to be more robust to outliers
+
+        def surface_eqn(data, a, b, c, d, f):
+            x, y = data
+            return a * x ** 2 + b * y ** 2 + c * x + d * y + f
+
+        # SETTINGS
+        # Note: Intial parameters for fitting set further down
+        r = 50  # Radius of square in pixels
```
ecobost: do you think these parameters will be changed at some point, maybe we could have a part table here called SurfaceParams with these params, that way if they ever change we will know what surfaces where computed each way. Although I guess if we are gonna change the params maybe we should just create a new SurfaceMethod altogether, up to you.

python/pipeline/stack.py
```diff
@@ -977,6 +978,129 @@ def make(self, key):
...
+        gaussian_blur_size = 5  # Size of gaussian blur applied to slice
+        min_points_allowed = 10  # If there are less than X points after filtering, throw an error
+        bounds = (
+        [0, 0, np.NINF, np.NINF, np.NINF], [np.Inf, np.Inf, np.Inf, np.Inf, np.Inf])  # Bounds for paraboloid fit
+        ss_percent = 0.40  # Percentage of points to subsample for robustness check
+        num_iterations = 1000  # Number of iterations to use for robustness check
+
+        print('Calculating surface of brain for stack', key)
+        full_stack = (stack.PreprocessedStack & key).fetch1('resized')
+        depth = full_stack.shape[0]
```
ecobost: `depth, height, width = full_stack.shape` looks a bit nicer

python/pipeline/stack.py
```diff
@@ -977,6 +978,129 @@ def make(self, key):
...
+
+                stack_slice_derivative = ndimage.sobel(blurred_slice)
+                surface_z = np.argmax(stack_slice_derivative)
+
+                if ((surface_z < upper_threshold_idx) and (blurred_slice[0] < z_0_upper_threshold) and
+                        (blurred_slice[-1] > 10)):
+                    surface_guess_map.append((surface_z, y, x))
+
+        if len(surface_guess_map) < min_points_allowed:
+            raise Exception(f"Surface calculation could not find enough valid points for {key}. Only {len(surface_guess_map)} detected")
```
ecobost: you could use PipelineException (it's already imported in stack), it's kind of the generic exception we use for pipeline errors. Advantage is that at least we know for sure that we are the ones raising that exception rather than Exception which could be raised by numpy or by some other package.

python/pipeline/stack.py
```diff
@@ -977,6 +978,129 @@ def make(self, key):
...
+
+        # Drop the z-values lower than 5th-percentile or greater than 95th-percentile
+        arr = np.array(surface_guess_map)
+        top = np.percentile(arr[:, 0], 95)
+        bot = np.percentile(arr[:, 0], 5)
+        surface_guess_map = arr[np.where(np.logical_and(arr[:, 0] > bot, arr[:, 0] < top))]
```
ecobost: minor: is the np.where needed there?

python/pipeline/stack.py
```diff
@@ -977,6 +978,129 @@ def make(self, key):
...
+                                       replace=False)
+            subsample = surface_guess_map[indices]
+            sub_popt, sub_pcov = curve_fit(surface_eqn, (subsample[:, 2], subsample[:, 1]), subsample[:, 0],
+                                           p0=initial, maxfev=10000, bounds=bounds)
+            all_sub_fitted_z[i, :, :] = surface_eqn((full_mesh_x, full_mesh_y), *sub_popt)
+
+        z_min_matrix = np.percentile(all_sub_fitted_z, 5, axis=0)
+        z_max_matrix = np.percentile(all_sub_fitted_z, 95, axis=0)
+
+        surface_key = {**key, 'surface_method_id': 1, 'surface_im': calculated_surface_map,
```
ecobost: surface method is gonna come with the key, no need to send it here. In fact, right at the start of the make you should probably check that surface_method is 1 and throw a NotImplementedError or PipelineException if it's not

Neuro-CS: I added a variable storing all valid method_ids and now raise a PipelineException if the key value is not in that list.

python/pipeline/stack.py
Neuro-CS: Nope. Fixed it.

python/pipeline/stack.py
Neuro-CS: Talking with Jake, I think we're going to make a new method if the settings need to be changed.

python/pipeline/stack.py
Neuro-CS: Not annoying at all. I changed it to fit that style.

python/pipeline/stack.py
```diff
@@ -1029,16 +1016,19 @@ def surface_eqn(data, a, b, c, d, f):
         upper_threshold_percent = 0.6  # Surface median intensity should be in the bottom X% of the *range* of medians
         gaussian_blur_size = 5  # Size of gaussian blur applied to slice
         min_points_allowed = 10  # If there are less than X points after filtering, throw an error
-        bounds = (
-        [0, 0, np.NINF, np.NINF, np.NINF], [np.Inf, np.Inf, np.Inf, np.Inf, np.Inf])  # Bounds for paraboloid fit
+        bounds = ([0, 0, np.NINF, np.NINF, np.NINF], [np.Inf, np.Inf, np.Inf, np.Inf, np.Inf])  # Bounds for paraboloid fit
         ss_percent = 0.40  # Percentage of points to subsample for robustness check
         num_iterations = 1000  # Number of iterations to use for robustness check
 
+        if 'surface_method_id' not in key:
```
ecobost: no need to check for this, surface_method_id will always be there

python/pipeline/stack.py
```diff
@@ -1029,16 +1016,19 @@ def surface_eqn(data, a, b, c, d, f):
...
+            raise PipelineException('Error: surface_method_id not specified')
+        if int(key['surface_method_id']) not in valid_method_ids:
```
ecobost: minor:not sure you need the int there

python/pipeline/stack.py
Neuro-CS: If I run populate and pass a stack key, how will the surface method be automatically filled out? Perhaps I'm missing something in my code.

ecobost: if you call Surface.populate(my_restr), the key passed to the make will be all in (PreprocessedStack * shared.SurfaceMethod & key).proj().fetch()

python/pipeline/stack.py
```diff
@@ -985,7 +986,8 @@ class Surface(dj.Computed):
     -> PreprocessedStack
     -> shared.SurfaceMethod
     ---
-    surface_im        : longblob          # Matrix of calculated depth for each pixel in stack
+    guessed_points    : blob              # Array of guessed depths stored in (z,y,x) format
```
ecobost: is blob big enough?

Neuro-CS: I guess I can error on the side of caution and make it a longblob.

PR #375, opened 2019-07-19
install pipeline from cajal, not local

DonnieKim411:


Dockerfile.tf-gpu
```diff
@@ -18,7 +18,7 @@ RUN git clone https://github.com/atlab/commons.git && \
     rm -r commons
 
 # Install pipeline
-COPY . /data/pipeline
-RUN pip3 install -e pipeline/python/
+RUN git clone https://github.com/cajal/pipeline.git && \
+    pip3 install pipeline/python
```
ecobost: I would leave the -e that allows for people to overwrite the pipeline code if they need to

PR #376, opened 2019-08-06
collection schema and stack.Area table

paulgfahey:
collection schema for experimenter organization of scans, curated and scored per suitability for a given study and scan/stack purpose

stack.Area for importing area maps from anatomy schema and registering into the stack 

meso.AreaMembership for assigning units into areas based on stack coordinates and stack.Area masks


python/pipeline/meso.py
```diff
@@ -1699,6 +1698,71 @@ def make(self, key):
             StackCoordinates.UnitInfo.insert1({**key, **unit_key, 'stack_x': unit_x,
                                                'stack_y': unit_y, 'stack_z': unit_z})
 
+anatomy = dj.create_virtual_module('pipeline_anatomy','pipeline_anatomy')
```
ecobost: this line should probably be moved all the way to the start of the file

PR #378, opened 2019-08-15
adding functions to read patch clamp ephys data. 

yizhoucc:


python/pipeline/utils/h5.py
```diff
@@ -94,6 +94,106 @@ def read_behavior_file(filename):
 
     return data
 
+def read_patchmaster(F):
+    if type(F)==str:
+        F=F[:-4]+'%d.h5' # chanage oneCell10_01.h5 to oneCell10_%d.h5
+        # Append '%d.h5' to filename in place of trailing 'x.h5'
+        # right now this way will only work for one digit file names, are we doing two digit names in the future? if so i suggest use the 0d format instead of having 9 and suddently 10 for eg.
```
ecobost: have you discussed this with someone? whoever wrote the code that writes the file will know if there will be more than one digit and also be in position to make changes to the way we record stuff

python/pipeline/utils/h5.py
```diff
@@ -94,6 +94,106 @@ def read_behavior_file(filename):
...
+    else:
+        raise Exception ('currently the input format as {} is not surrpoted, please use str'.format(type(F)))
+    f=h5py.File(F,"r+",driver='family',memb_size=0) # read only, mimic if we are reading a batch of files, but the batch only have 0 member size
+    #begin case 7
+    '''
```
ecobost: if this is documentation you can put it under read_patchmaster(F):
```python
def read_patchmaster(F):
    """ Read patch clamp data recorded with ....

    < More info about the function (if any) >

     :param string F: Filename .......

     :returns: ......
    """
```
check `read_behavior_file` in this same file for an example

python/pipeline/utils/h5.py
```diff
@@ -94,6 +94,106 @@ def read_behavior_file(filename):
 
     return data
 
+def read_patchmaster(F):
+    if type(F)==str:
+        F=F[:-4]+'%d.h5' # chanage oneCell10_01.h5 to oneCell10_%d.h5
+        # Append '%d.h5' to filename in place of trailing 'x.h5'
```
ecobost: usually the comments describing what you are doing go before the line:
```python3
# This describes the following block of code
skslsl
slslslls

# this describes the next block of code
slslskjfsj
sksslsl 
slslslslsl   # specific comment for this tricky line
```
so it this case it will be 
```python
if ....
    # Append '%d.h5' to filename in place of trailing 'x.h5'
    F = F[:-4] + '%d.h5'
else: 
    skslls
````

File: python/pipeline/utils/h5.py [hunk 4]
```diff
@@ -94,6 +94,106 @@ def read_behavior_file(filename):
...
+        print('v low pass gain changed')	
+        raise Exception('vlowpasscorner changed ')
+
+    vHighPassCorners = [0, 0.1, 0.3, 0.5, 1, 3, 5, 10 ,30 ,50 ,100, 300, 500, 800 ,1000, 3000]
+    if np.unique(sets[4,:].round()).size==1:
+        settings1['vHighPass']=vHighPassCorners[int(round(np.unique(sets[4,:])[0]))-1+9]
+    else:
+        print('v high pass gain changed')	
+        raise Exception('vhighpasscorner changed ')
+    '''
```
ecobost: you can probably drop this code

python/pipeline/utils/h5.py
```diff
@@ -94,6 +94,106 @@ def read_behavior_file(filename):
...
+        settings1['vLowPass']=vLowPassCorners[int(round(np.unique(sets[3,:])[0]))-1+9]
+    else:
+        print('v low pass gain changed')	
+        raise Exception('vlowpasscorner changed ')
+
+    vHighPassCorners = [0, 0.1, 0.3, 0.5, 1, 3, 5, 10 ,30 ,50 ,100, 300, 500, 800 ,1000, 3000]
+    if np.unique(sets[4,:].round()).size==1:
+        settings1['vHighPass']=vHighPassCorners[int(round(np.unique(sets[4,:])[0]))-1+9]
+    else:
+        print('v high pass gain changed')
```
ecobost: no need to print and send an exception, just raise the Exception

python/pipeline/utils/h5.py
```diff
@@ -94,6 +94,106 @@ def read_behavior_file(filename):
...
     return data
 
+def read_patchmaster(F):
+    if type(F)==str:
+        F=F[:-4]+'%d.h5' # chanage oneCell10_01.h5 to oneCell10_%d.h5
+        # Append '%d.h5' to filename in place of trailing 'x.h5'
+        # right now this way will only work for one digit file names, are we doing two digit names in the future? if so i suggest use the 0d format instead of having 9 and suddently 10 for eg.
+    else:
+        raise Exception ('currently the input format as {} is not surrpoted, please use str'.format(type(F)))
+    f=h5py.File(F,"r+",driver='family',memb_size=0) # read only, mimic if we are reading a batch of files, but the batch only have 0 member size
```
ecobost: usually better to read files in a with statement
``` python
with f as h5py.File(F, 'r+', driver='family', memb_size=0):
    # use f
```
that takes care of closing the file if something breaks in the middle of the with and when you leave the function

python/pipeline/utils/h5.py
```diff
@@ -94,6 +94,106 @@ def read_behavior_file(filename):
...
+    data['v1']=wf[1,:]
+    data['i2']=wf[7,:]
+    data['v2']=wf[8,:]
+    data['syncPd']=wf[2,:]
+    data['stimPd']=wf[3,:]
+    data['led']=wf[4,:]
+    data['command']=wf[5,:]
+    data['shutter']=wf[6,:]
+    data['scanImage']=wf[9,:]
+    data['ts']=wf[10,:]
```
ecobost: can you run the formatting tool in your code editor. in pycharm, it is Ctrl+Alt+L. That will add empty lines when necessary, add spaces before and after the =, spaces before the commas, break lines that are too long, and other stuff like that

PR #382, opened 2019-09-27
Adding odor pipeline schema

Neuro-CS:
Adds a new odor.py file for olfaction pipeline, updates h5.py with related functions, and fixes meso SegmentationManual.m being empty.

### DIFF
--- a/matlab/schemas/+meso/SegmentationManual.m
+++ b/matlab/schemas/+meso/SegmentationManual.m
@@ -6,12 +6,91 @@
 
 classdef SegmentationManual < dj.Computed
 
-	methods(Access=protected)
+%{
+# masks created manually
+-> meso.Segmentation
+---
+%}
 
+
+	classdef SegmentationManual < dj.Computed
+
+	    properties
+		popRel  = meso.SummaryImagesAverage
+	    end
+
+	    methods(Access=protected)
 		function makeTuples(self, key)
-		%!!! compute missing fields for key here
-% 			 self.insert(key)
+		    images = ne7.mat.normalize(fetch1(meso.SummaryImagesAverage & key, 'average_image'));
+
+		    % add second channel info
+		    key2 = key;
+		    channels = [1 2];
+		    key2.channel = channels(key.channel~=channels);
+		    if exists(meso.SummaryImagesAverage & key2);
+		       images = repmat(images,1,1,3);
+		       image2 = ne7.mat.normalize(fetch1(meso.SummaryImagesAverage & key2, 'average_image'));
+		       images(:,:,:,2) = repmat(image2,1,1,3);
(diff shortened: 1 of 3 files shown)


python/pipeline/utils/h5.py
```diff
@@ -234,4 +234,72 @@ def find_flips(signal, fps, monitor_fps):
...
+
+
+def read_digital_olfaction_file(filename):
+    """ Reads hdf5 files with olfaction digital signals.
+    :param filename: path of the file. Needs a %d where 2GB file split counter is located.
+    :returns: A dictionary with these fields:
+        version: string. File version
+        delay: float. Time in milliseconds between puffs of odor
+        duration: float. Duration of odor presentation in milliseconds.
+        valves: 1-d array (num_samples). State of valve stored as decimal of binary
```
ecobost: +1 for documentation :)

matlab/schemas/+meso/SegmentationManual.m
```diff
@@ -5,13 +5,81 @@
 
 
 classdef SegmentationManual < dj.Computed
+    
+    properties
+        popRel  = meso.SummaryImagesAverage
+    end
+    
+    methods(Access=protected)
+        function makeTuples(self, key)
```
ecobost: Don't really know whether this is ok, but I guess it matches the reso MannualSegmentation code

python/pipeline/odor.py
```diff
@@ -0,0 +1,268 @@
...
+import datajoint as dj
+import numpy as np
+
+from pipeline import mice
+from pipeline import meso
+from pipeline.utils import h5
+from commons import lab
+
+schema = dj.schema('pipeline_odor', locals(), create_tables=False)
+CURRENT_VERSION = 1
```
ecobost: you are not using the version so you may not need this. and if you are, it's probably better to have a version (as secondary attribute) per table rather than one version as PK over the entire schema

python/pipeline/odor.py
```diff
@@ -0,0 +1,268 @@
...
+        # Finally, reverse list to make first index encode first valve
+        valve_states = [bool(int(d)) for d in format(int(num), 'b')][::-1]
+        return np.array(valve_states)
+
+    def make(self, key):
+
+        print(f'Populating trials for {key}')
+
+        # Get olfactory h5 path and filename
+        olfactory_path = (OdorSession & key).fetch1('odor_path')
```
ecobost: https://github.com/cajal/pipeline/blob/master/python/pipeline/experiment.py#L467 In experiment.session we have the attribute local_filenames_as_wildcard that takes care of finding the path so we don't have to rewrite it every time, maybe you can have something similar. Definitely not needed , if you think this works better for this pipeline.

PR #385, opened 2019-10-18
Updating pupil related stuff + new tables: Projector and ProjectorDisplay

DonnieKim411:
1. Get rid of `filter_config`
2. remove `tracking` directory and its sub-directories
3. method name changed from `DLC_tools.filter_by_std` to `DLC_tools.filter_by_fitting_std`
4. delete `OnlineMedianFilteredFittedPupil` table (it is already dropped on DB side
5. missing schema decorator on `SurfaceMethod` table
6. bug fixed when finding cropping coords from short video (DLC related bug) (commit 
`a861127`)
7. Projector and ProjectorDisplay tables added


matlab/schemas/+experiment/Projector.m
```diff
@@ -0,0 +1,15 @@
+%{
+experiment.Projector (lookup) #
```
ecobost: how come this first line does not crash, is it assumed that the first line will be a comment? (can't find docs on https://docs.datajoint.io/matlab/definition/03-Table-Definition.html). if it is a comment it needs to be better than `experiment.Projector(lookup)` that doesn't add any information

matlab/schemas/+experiment/ProjectorDisplay.m
```diff
@@ -0,0 +1,12 @@
+%{
+experiment.ProjectorDisplay (Manual) #
```
ecobost: same here for the first line not being very descriptive

python/pipeline/experiment.py
```diff
@@ -593,4 +593,35 @@ class AutoProcessing(dj.Manual):
...
     """
 
+@schema
+class Projector(dj.Lookup):
+    definition = """
+    # projector specifications
+    projector_id        : tinyint                               # projector id
+    refresh_rate        : tinyint                               # refresh rate in Hz
+    ---
+    red                 : enum('UV', 'Green', 'Blue', 'None')   # color to be used for red channel
```
ecobost: I generally don't like enums because they are hard to extend. Consider adding a lookup table with colors and use that here
```
-> Color.proj(red_color='color') # color to be used for red channel
```

matlab/schemas/+experiment/Projector.m
```diff
@@ -0,0 +1,15 @@
+%{
+experiment.Projector (lookup) # 
+projector_id        : tinyint                       # projector id
+---
+pixel_width         : smallint                      # number of pixels in width
+pixel_height        : smallint                      # number of pixels in height
+red                 : enum('UV', 'Green', 'None')   # color to be used for red channel
```
ecobost: this enum does not match the enum in .py

python/pipeline/experiment.py
```diff
@@ -593,4 +593,35 @@ class AutoProcessing(dj.Manual):
     autosegment=false   :boolean       # segment somas in the first channel with default method
     """
 
+@schema
+class Projector(dj.Lookup):
+    definition = """
+    # projector specifications
+    projector_id        : tinyint                               # projector id
+    refresh_rate        : tinyint                               # refresh rate in Hz
```
ecobost: can the refresh rate not be a float?
and if the refresh rate is different is it still the same projector?, bit odd to see the refresh rate (which I thought would be a dependent attribute of the projector as primary key)

python/pipeline/experiment.py
```diff
@@ -593,4 +593,35 @@ class AutoProcessing(dj.Manual):
...
+        [1, 30, 'UV', 'UV', 'Green', 1140, 912]
+    ]
+
+@schema
+class ProjectorDisplay(dj.Manual):
+    definition = """
+    # projected display specification
+    -> Projector
+    -> Rig
+    target_distance     : decimal(6,3)  # distance from mouse to the projected display in cm
```
ecobost: you sure target distance should be in the primary key

python/scripts/pupil-data-copy.py
```diff
@@ -0,0 +1,39 @@
+from pipeline import pupil
```
ecobost: can you make a little docstring comment at the top of this file saying what it does: "Bringing from old whatever whatever table to new xxx table" or something like that

matlab/schemas/+experiment/Projector.m
DonnieKim411: it turned out that the way i did was an outdated style of defining tables. It still works but i changed so that it meets the latest style. fixed

matlab/schemas/+experiment/ProjectorDisplay.m
DonnieKim411: same as above

python/pipeline/experiment.py
DonnieKim411: I think using enum is fine... if we need to extend later, one can easily go back to SQL and add a color if needed...

matlab/schemas/+experiment/Projector.m
DonnieKim411: fixed

python/pipeline/experiment.py
DonnieKim411: refresh rate is already in int? not a float?

Refresh rate can be _adjusted_ on the hardware level settings. So by specifying it as a primary key, we ensure that the refresh rate that was set by the project (physical level) must match with which projector AND which refresh setting (software level).

python/pipeline/experiment.py
DonnieKim411: yes, this is a primary key because the distance between the mouse and the display later is used for grating generation process https://github.com/DonnieKim411/stimuli/blob/master/matlab/%2Bstimulus/ColorGrate.m#L69

depending on the target_distance, the way grating is visualized on display changes, hence it is a primary key

python/scripts/pupil-data-copy.py
DonnieKim411: Should I just version this whole module instead and make a document on it?

python/pipeline/experiment.py
ecobost: can you? at some point I was told changing enums will be annoying. Even if it's doable, I doubt it is recommended relying on altering the table structure when new changes come along, using enum is reasonable when there is no reasonable thought that it will be extended later.

python/pipeline/experiment.py
ecobost: I was asking whether you are sure refresh rate will never need to be a float

python/scripts/pupil-data-copy.py
ecobost: not sure what is the "whole module"? I thought a comment would be ok but if you wanna do something nelse go ahead

python/pipeline/experiment.py
ecobost: that still seems like something that should be documented but not as part of the primary key. It seems class Projector is doing too many things, both recording different types of projector (whose key will be projector id and should have projector properties experimenters think are improtant, like maker of the projector, when we acquired it, how rgb inputs map to projected colors etc) and different configurations of that projector (which right now is just refresh_rate but may need more stuff to be saved and hopefully is projector independent), you could separate it into different classes. why do that, because if I have a table that depends on the projector but not its particular configuration I couldn't point to Projector right now becuase that will state that it depends on an specific refresh_rate too

python/scripts/pupil-data-copy.py
DonnieKim411: https://github.com/cajal/pipeline/pull/385/commits/c28e9ecdc1fe2f1099b345e2b5124212a4989602

This is what i meant

python/pipeline/experiment.py
ecobost: here's the theoretical reason why it seems odd: http://agiledata.org/essays/dataNormalization.html It't not normalized (it's only 1NF, not 2NF as non-dependent attributes do not depend on the refresh rate)

DonnieKim411: oh. well i can change to `decimal` then

python/pipeline/experiment.py
ecobost: i don't understand what this table is supposed to contain. I don't see how the rig matters in defining the size of displayed width for instance, if a prohjector witht he same config equally far is in rig 1 or 2 that shouldn't matter. Also if target distance matters for creating the grating it should be part of some GratingConfigs table (or is this the GratingConfig table but with a diff name?)

ecobost: in a less technichal note I doubt very much that experimetners will record how far the mouse is to the screen for each scan, nor will it be very exact, chances are this should just be assumed in the condition

python/pipeline/experiment.py
dimitri-yatsenko: refresh_rate should be a secondary attribute, not in primary key

python/pipeline/experiment.py
dimitri-yatsenko: `target_distance` is not a primary attribute, should be moved to secondary

dimitri-yatsenko: Can multiple projectors be used on the same rig? If not, then we don't need both in the primary key.

python/pipeline/experiment.py
```diff
@@ -593,23 +593,37 @@ class AutoProcessing(dj.Manual):
...
+    definition = """
+    # color options for projector channels
+    color_id        : tinyint     # color id
+    ---
+    color           : varchar(32) # color name      
+    """
+    contents = [[0, 'none'],
+                [1, 'red'],
+                [2, 'green'],
+                [3, 'Blue'],
```
ecobost: can you change this to lowercase blue

python/pipeline/experiment.py
```diff
@@ -593,23 +593,37 @@ class AutoProcessing(dj.Manual):
...
 class Projector(dj.Lookup):
     definition = """
     # projector specifications
     projector_id        : tinyint                               # projector id
-    refresh_rate        : tinyint                               # refresh rate in Hz
     ---
-    red                 : enum('UV', 'Green', 'Blue', 'None')   # color to be used for red channel
-    green               : enum('UV', 'Green', 'Blue', 'None')   # color to be used for green channel
-    blue                : enum('UV', 'Green', 'Blue', 'None')   # color to be used for blue channel
+    red                 : tinyint                               # color to be used for red channel
```
ecobost: I thought this should point to ProjectorColor

ecobost: nvm, somehow github was making me review old code

python/pipeline/ProjectorColor.m
```diff
@@ -0,0 +1,9 @@
+%{
```
ecobost: I think the name of the class should agree with the name of the file. But also, I don't think this table should be here. It is already under pipeline/experiment.py

python/scripts/pupil-data-copy.py
ecobost: oh ok

ecobost: actually I may still add a docstring in this file pupil-data-copy.py just to say whta it does. Literally just a line so people can call help on it and see what it does or so they don't have to read teh code

python/pipeline/ProjectorColor.m
ecobost: still unresolved (we either need to put this in experiment or in pipeline, I think it makes more sense in experiment as all other projector stuff is there)

python/pipeline/experiment.py
```diff
@@ -593,4 +593,66 @@ class AutoProcessing(dj.Manual):
     autosegment=false   :boolean       # segment somas in the first channel with default method
     """
 
+@schema
+class ProjectorColor(dj.Lookup):
```
ecobost: this does not agree with the ProjectorColor.m file

ecobost: I think you need to modify it here and probably change the contents of the table

python/pipeline/pupil.py
```diff
@@ -1,3 +1,18 @@
+"""
+Date: 10/29/2019
+This module is for pupil related tables and computation.
+Current version is at 1.0.0 
+
+What's changed in this version:
```
ecobost: maybe you can add the date of the current version to make it easier to search later on Github

python/scripts/pupil-data-copy.py
ecobost: still unresolved

python/pipeline/pupil.py
DonnieKim411: updated

python/scripts/pupil-data-copy.py
DonnieKim411: added the docstring under pupil-data-copy.py specifying what this job did

python/pipeline/ProjectorColor.m
DonnieKim411: deleted. Idk why it was still there

python/pipeline/experiment.py
DonnieKim411: Updated

PR #386, opened 2019-11-13
make pipeline importable without cv2 installed

dimitri-yatsenko:


ecobost:
We still use cv2 (and import it directly) in pipeline/movies.py, pipeline/pupil.py and pipeline/posture.py. i think I would rather get rid of the cv2 check during import here and just state that if you don't have opencv you won't be able to import eye_tracking (in the same way you won't be able to import any of the other ones)

PR #389, opened 2019-12-20
Added K8 jupyter notebook deployment and update minions yaml files to…

Synicix:
… follow standard design


Synicix:
I added some folders. One is cronjobs which you can just do kubectl apply -f cronjobs and it will deploy everything in the folder. The other folder is for the jupyter notebook

PR #390, opened 2020-02-14
Added motion correct passthrough for high FPS

Neuro-CS:
This is a fix for voltage imaging done through reso.

### DIFF
--- a/python/pipeline/reso.py
+++ b/python/pipeline/reso.py
@@ -9,6 +9,7 @@ from . import experiment, notify, shared
 from .utils import galvo_corrections, signal, quality, mask_classification, performance
 from .exceptions import PipelineException
 
+
 schema = dj.schema('pipeline_reso', locals(), create_tables=False)
 CURRENT_VERSION = 1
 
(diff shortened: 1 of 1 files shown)


python/pipeline/reso.py
```diff
@@ -404,56 +404,69 @@ def make(self, key):
...
+            # Center shifts around zero
+            y_shifts -= np.median(y_shifts)
+            x_shifts -= np.median(x_shifts)
+
+        else:
+
+            # Fill with dummy values
+            print(f'FPS of {int(fps)} is greater than 100. Not correcting for motion.')
+            y_shifts = np.zeros(scan.num_frames)
+            x_shifts = np.zeros(scan.num_frames)
```
ecobost, line 466: can you make this dtype np.float32

python/pipeline/reso.py
```diff
@@ -404,56 +404,69 @@ def make(self, key):
...
+            outliers = [False]*scan.num_frames
```
ecobost, line 467: you can also do np.zeros(scan.num_frames, dtype=boolean)

python/pipeline/reso.py
```diff
@@ -404,56 +404,69 @@ def make(self, key):
         scan = scanreader.read_scan(scan_filename)
```
ecobost, line 404: if you are not gonna perform correction then there is no need to read the scan

PR #399, opened 2020-04-27
Add voltage imagining pipeline

Neuro-CS:
Create vreso.py to contain classes for voltage imaging and ephy. This is a rough draft, but necessary for collaborative analysis at the moment. It will be expanded upon in the future.


python/pipeline/vreso.py
```diff
@@ -0,0 +1,175 @@
...
+schema = dj.schema("pipeline_vreso", locals(), create_tables=False)
+CURRENT_VERSION = 1
+
+
+@schema
+class Amplifier(dj.Lookup):
+    definition = """ # Electrophysiology rig setup
+    amplifier                     : varchar(32)         # name and model of patch rig
+    """
+    contents = [
```
ecobost, line 21: we usually add a description field (not as primary key) with some text about what the entry means. Unless the name is more than enough to describe it (like "red")

python/pipeline/vreso.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+@schema
+class Recording(dj.Manual):
+    definition = """ # Single recording of a cell
+    -> Cell
+    recording_id                  : smallint unsigned   # recording index for patch session
+    ---
+    clamp_type                    : enum('voltage','current')
+    patch_type                    : enum('whole', 'loose')
```
ecobost, line 58: enums are generally discouraged because if you wanna add more values later, you would have to go to MySQL (and I thought it was not even possible but i just googled it and i think it is); if you don;t think there will be ny new types of clamp or patch, should be alright, otherwise you could add PatchType as a lookup (added benefit is that this lookup can have other properties like a description of what it means)

python/pipeline/vreso.py
```diff
@@ -0,0 +1,175 @@
...
+    file_name = ''                : varchar(128)        # filename of h5 file containing recording data
+    patch_ts=CURRENT_TIMESTAMP    : timestamp           # time of recording
+    """
+
+
+@schema
+class RecordingInfo(dj.Imported):
```
ecobost, line 65: not sure why Imported rather than Computed? make still works?

python/pipeline/vreso.py
```diff
@@ -0,0 +1,175 @@
...
+    definition = """ # Imported information of a single recording of a cell
+    -> Recording
+    ---
+    voltage                       : external            # (V_amplifier) recorded voltage
```
ecobost, line 69: how big are these, even as a external if they are dozens of GBs per cell may be worth asking whether it is ok to store them again (original is in the h5 i assume)

python/pipeline/vreso.py
```diff
@@ -0,0 +1,175 @@
...
+            binarized_valid = binarized_signal[rising_edges[0]: rising_edges[-1]][valid_times]
+            frame_duration = np.mean(binarized_valid) * frame_period
+            falling_edges = np.where(np.diff(binarized_signal.astype(int)) < 0)[0]
+            last_frame_duration = patch_times[falling_edges[-1]] - frame_times[-1]
+            if (np.isnan(last_frame_duration) or last_frame_duration < 0 or
+                    abs(last_frame_duration - frame_duration) > 0.15 * frame_duration):
+                frame_times = frame_times[:-1]
+
+            ####### WARNING: NO CORRECTION APPLIED - TEMP CODE #######
+            voltage = f['waveform'][1, :]
```
ecobost, line 144: make sure this are float32 (or int8, int16, int32, anything other than 64 bit numbers)

python/pipeline/vreso.py
```diff
@@ -0,0 +1,175 @@
...
+            current = f['waveform'][0, :]
+            command = f['waveform'][5, :]
+
+            ####### WARNING: DUMMY VARIABLES - TEMP CODE #######
+            vgain = 0
+            igain = 0
+            command_gain = 0
+
+            h5py.File.close(f)
```
ecobost, line 153: you don't need this line that's what the with open() as f statement does

python/pipeline/vreso.py
```diff
@@ -0,0 +1,175 @@
+import scanreader
+import numpy as np
+import datajoint as dj
+from pipeline.exceptions import PipelineException
+from pipeline import reso, experiment, mice, notify, shared
+from pipeline.utils import galvo_corrections, signal, quality, mask_classification, performance, h5
```
ecobost, line 6: a lot of these are not used in the code, importing them makes people believe they are

python/pipeline/vreso.py
```diff
@@ -0,0 +1,175 @@
+import scanreader
```
ecobost, line 1: vreso stands for voltage resonant, i thought this was other voltage reconrdings

python/pipeline/vreso.py
Neuro-CS: Since this schema a rough draft, these libraries will be used later. I removed parts that are incomplete.

python/pipeline/vreso.py
Neuro-CS: There are only two types of clamps possible and patching is the same.

python/pipeline/vreso.py
```diff
@@ -0,0 +1,175 @@
...
+from pipeline.exceptions import PipelineException
+from pipeline import reso, experiment, mice, notify, shared
+from pipeline.utils import galvo_corrections, signal, quality, mask_classification, performance, h5
+
+# external storage place
+dj.config['external'] = {'protocol': 'file',
+                         'location': '/mnt/dj-stor01/pipeline-externals'}
+
+schema = dj.schema("pipeline_vreso", locals(), create_tables=False)
+CURRENT_VERSION = 1
```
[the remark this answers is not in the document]
Neuro-CS: I had an error recently in a reso insert command due to not having CURRENT_VERSION defined. Are you sure it's unnecessary?

python/pipeline/vreso.py
Neuro-CS: These are ~90-200MB, so it should be fine.

PR #401, opened 2020-06-15
Update K8 deployment yaml files

Synicix:


K8/CronJobs/minion-gpu.yaml
```diff
@@ -3,7 +3,7 @@ kind: CronJob # This tells kubernetes what kind of class it is working with
 metadata:
   name: minion-gpu
 spec:
-  schedule: "*/30 * * * *" # every hour
+  schedule: "*/5 * * * *" # every hour
```
ZhuokunDing, line 6: The comment looks inconsistent with the spec

Synicix: Ah yeah let me fix that

Synicix: Yeah these was cobos comments when I was working on creating this template file

PR #406, opened 2020-09-11
Add FieldAnnotation table and Injection schema

Neuro-CS:
- FieldAnnotation added to meso.py & reso.py
- shared.py updated to include FieldAnnotation lookup (ExpressionConstruct)
- injection.py schema added


jakereimer (approved):
A few wording edits in Injection schema, but ready to merge.

PR #418, opened 2021-03-22
Add clocktools, pupil processing, treadmill processing

Neuro-CS:
Three major changes:

1. New file in pipeline.utils called clocktools.py
     - Adds auto-conversion functions between any clocks (behavior, stimulus, odor)
     - Adds auto-interpolation and fetching of signals for nearly all signals of interest (fluorescence, deconvolved activity, respiration, treadmill, pupil)
     - Takes as input indices or time boundaries on any clock

2. New tables to pupil.py
     - ProcessedPupil auto-filters pupil radii
     - Other tables store all detected dilation/constriction events or define methods to do so

3. New tables to treadmill.py
     - Running table detects bouts of running and stores each event
     - RunningMethod stores different methods to detect running like threshold + max gap between the same running bout

### DIFF
--- a/python/pipeline/pupil.py
+++ b/python/pipeline/pupil.py
@@ -1348,11 +1348,11 @@ class PupilPeriods(dj.Computed):
         -> PupilPeriods
         period_idx                : smallint unsigned
         ---
-        period_onset              : float               # period start time (in seconds)
-        period_offset             : float               # period stop time (in seconds)
-        peak_change_time          : float               # time (in seconds) at which peak change rate occurs
-        period_delta              : float               # overall change in size (in pixels, as a placeholder for now)
-        period_duration           : float               # period duration (in seconds)
+        period_onset              : float               # (sec) period start time
+        period_offset             : float               # (sec) period stop time
+        peak_change_time          : float               # (sec) time at which peak change rate occurs
+        period_delta              : float               # (mm) overall change in size (dilation > 0, constriction < 0)
+        period_duration           : float               # (sec) period duration
         """
         
     def _make_tuples(self, key):
--- a/python/pipeline/pupil.py
+++ b/python/pipeline/pupil.py
@@ -32,6 +32,7 @@ import datajoint as dj
 from datajoint.jobs import key_hash
 from datajoint.autopopulate import AutoPopulate
 
+from .utils import clocktools
 from .utils.decorators import gitlog
 from .utils import eye_tracking, h5
 from .utils.eye_tracking import PupilTracker, ManualTracker
@@ -43,6 +44,7 @@ from IPython import display
 import pylab as pl
 import matplotlib.pyplot as plt
 from matplotlib.patches import Ellipse
+from scipy import interpolate
 import time
 import datetime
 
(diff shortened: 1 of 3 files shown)


python/pipeline/utils/clocktools.py
```diff
@@ -0,0 +1,996 @@
...
+    ## Convert source indices to time boundaries, then convert time boundaries into target indices
+    ##
+
+    ## Convert indices into start/end times for each continuous fragment (incrementing by 1)
+    time_boundaries = find_time_boundaries(indices, source_times, drop_single_idx)
+    target_indices = []
+    single_idx_count = 0
+
+    ## Loop through start & end times and create list of indices corresponding to that block of time
+    for [start, end] in time_boundaries:
```
ramosaj, line 511: would be really nice if these next ~20 lines of code could be put into a function, that way you could reuse them all across the convert_clocks_*_to_*

python/pipeline/utils/clocktools.py
```diff
@@ -0,0 +1,996 @@
...
+import datajoint as dj
+from scipy import interpolate
+from stimulus import stimulus
+from pipeline.exceptions import PipelineException
+from pipeline import treadmill, fuse, shared, odor
+
+pupil = dj.create_virtual_module("pipeline_eye", "pipeline_eye")
+
+
+def find_idx_boundaries(indices, drop_single_idx=False):
```
ramosaj, line 12: I think there is a way to do this with ittertools.groupby, would save on a lot of code and documentation. [Here ](https://docs.python.org/3/library/itertools.html#itertools.groupby)is the documentation

python/pipeline/utils/clocktools.py
```diff
@@ -0,0 +1,996 @@
...
+                                 'deconvolution-behavior', 'pupil', 'treadmill', 'respiration'
+
+                target_type: A string specifying what indices to convert into. Fluorescence and deconvolution have a 
+                             dash followed by "behavior" or "stimulus" to refer to which clock you are using.
+                             Supported options: 
+                                 'fluorescence-stimulus', 'deconvolution-stimulus', ,'fluorescence-behavior',
+                                 'deconvolution-behavior', 'pupil', 'treadmill', 'respiration'
+
+                return_interpolate: Boolean with the following behavior
+                                     - True: Interpolate source signal onto target clock, return interpolated signal
```
ramosaj, line 447: _source_ signal or _target_ signal?

python/pipeline/utils/clocktools.py
```diff
@@ -0,0 +1,996 @@
...
+    signal_interp = interpolate.interp1d(
+        source_times, source_signal, bounds_error=False
+    )
+    interpolated_signal = signal_interp(target_times)
+    if target_replace_nans is not None:
+        for target_nan_idx in target_replace_nans:
+            interpolated_signal[target_nan_idx] = np.nan
+
+    return interpolated_signal
+
```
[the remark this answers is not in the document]
paulgfahey: I think these four functions could be collapsed into a single one, where source format = ('index' , 'time') and desired target format = ('index' , 'time' , 'signal') was explicitly specified in the calling.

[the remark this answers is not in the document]
paulgfahey: I was also really confused that return_interpolate would result in returning source signal rather than target signal.  I would prefer that the output was always from the target class, otherwise the source target nomenclature becomes muddied very quickly.  If the source signal is required, the function can be called again with the roles reversed.  

Is there a nuanced use case where it makes sense to return the source signal in this way instead?

python/pipeline/utils/clocktools.py
```diff
@@ -0,0 +1,996 @@
...
+        if len(source_signal_fragments) == 1:
+            source_signal_fragments = source_signal_fragments[0]
+
+        return source_signal_fragments
+
+    else:
+
+        return target_indices
+    
+
```
ramosaj, line 556: All of the above comments also refer to the rest of the convert_clocks_*_to_*, since they all follow pretty much the same documentation/outline.

python/pipeline/pupil.py
```diff
@@ -1151,6 +1153,352 @@ def plot_fitting(key, start, end=-1, fit_type='Circle', fig=None, ax=None, mask_
...
+    -> PupilPeriodMethod
+    -> PupilUnitConversion
+    ---
+    dilation_freq                 : float               # dilations per second (within valid start/end)
+    constriction_freq             : float               # constrictions per second (within valid start/end)
+    num_pupil_periods             : smallint unsigned   # total number of dilation and constriction periods
+    """
+
+    class DilationConstriction(dj.Part):
+        definition = """ # Single period of dilation or constriction
```
ramosaj, line 1347: does the period_index specify whether it's dilation or constriction? if so, the comment should say that. If not, how do you tell the difference between a constriction and dilation period?

python/pipeline/pupil.py
```diff
@@ -1151,6 +1153,352 @@ def plot_fitting(key, start, end=-1, fit_type='Circle', fig=None, ax=None, mask_
...
+            px_to_mm = (PupilUnitConversion & key).fetch1('conversion_factor')
+
+            ## Get timing data
+            pupil_times = (Eye & key).fetch1('eye_time')
+            pupil_fps = 1/np.nanmedian(np.diff(pupil_times))
+
+            ## Calculate peaks and valleys over nan-interpolated radii
+            ## np.diff(np.sign(derivative)) creates an array that's -1 when at a peak and 1 at valleys
+            ## (Apologies in advance for that opaque one-liner)
+            pupil_radii = (ProcessedPupil & key).fetch1('filtered_pupil_radius')
```
ramosaj, line 1380: Since this is used below in plot_pupil_periods, it might be a good idea to have this as its own helper function.

python/pipeline/treadmill.py
```diff
@@ -95,4 +97,252 @@ def notify(self, key):
...
+        """
+
+        ## Splits list of indices into a list of lists. Each sublist contains indices which continuously increase by 1.
+        ## This function also drops sublists which only contain a single idx.
+        running_fragments = clocktools.find_idx_boundaries(running_indices, drop_single_idx=True)
+
+        ## If we have two or more fragments, test if any should be combined
+        if len(running_fragments) > 1:
+
+            ## Basic idea: Have a temporary start index. Loop through all lists of indices. Whenever there is a list
```
ramosaj, line 233: I'm confused about what this is doing. So, if there is a long enough gap between running periods then they are combined? Shouldn't it be the other way around, i.e. if the gap is short enough then combine them?

Neuro-CS: You can think of it like having two cursors, one for start_idx and another for end_idx. Start by setting start and end cursors on the start and end of the first fragment. When looking at the second fragment, one of two things happens:
1. The gap between the first and second fragment is below threshold. In that case, set the end cursor to the end of the second fragment and keep looping through.
2. The gap between the first and second fragment is greater than the threshold. In that case store an np.arange from the start cursor to the end cursor before setting those cursors to the start and end of the second fragment, then continue looping through fragments.

Does that seem reasonable to you?

ramosaj (approved):
After all the changes this is ready to merge!

PR #423, opened 2021-04-28
Added OdorSync fix for odd scans

Neuro-CS:
Added error checking if irregular frame times are detected. Code attempts to find longest fragment and fix the error automatically, if possible.


jakereimer (commented):
Maybe I am missing something in the elif condition, but if the longest fragment equals nFrames, it need not be the case that there is a single additional error frame at the beginning. I think this should be tested explicitly or the fix should be made more general.

jakereimer (approved):

PR #445, opened 2022-03-02
Update h5.py

paulgfahey:
THIS IS A HOTFIX THAT SHOULD BE REPLACED

h5 util previously failed to update channel indices where channels were grouped under 1:x notation for x channels of the same type -- added string comprehension loop to replace that with multiple channel names so indexing could proceed normally.

### DIFF
--- a/python/pipeline/utils/h5.py
+++ b/python/pipeline/utils/h5.py
@@ -78,6 +78,13 @@ def read_behavior_file(filename):
             analog_signals = np.array(f['Analog Signals'])
             channel_names = f.attrs['AS_channelNames'].decode('ascii').split(',')
             channel_names = [cn.strip() for cn in channel_names]
+            
+            for i,channel_name in enumerate(channel_names): # only works up to 9 channels
+                if ':' in channel_name:
+                    colon_index = channel_name.find(':')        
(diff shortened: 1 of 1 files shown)


saumilpatel (approved):

paulgfahey:
To elaborate for the curious:
During the experiments on 2P3 today, we realized that the additional ephys channel that is not normally present on other scans (on 2P3 or on meso) was encoded in the h5 file in a way that broke the indexing of different traces of different types, including the time stamps.  As a result, stimulus.Sync and any other code dependent on the timestamps in the h5 file were unable to run.  The solution was to expand the notation of the additional ephys channels such that they occupied the appropriate number of entries in the header -- going from "Ephys1:2" in one line to "Ephys1" and "Ephys2" in two separate lines.  Thus, indexing by the header entries successfully pointed to the corresponding trace.  This change was made in the pipeline utils file, so anything pulling the h5 files and using the h5.read_behavior_file utility will be able to correctly extract the timestamps at the appropriate index after this change.  This change was made by Paul in discussion with Saumil, who added the new Ephys channel and was thus familiar with the impact.  It was reviewed and accepted by Saumil.  It is only expected to affect analysis of scans with the additional Ephys channels in them (the sleep studies being run on 2p3 this week) using h5 components.

PR #448, opened 2022-06-15
Added copydata to pipeline.utils

Neuro-CS:
copydata.py is a module used to transfer data from any two DataJoint servers, but has multiple specialized functions to copy data from the Tolias database to the Reimer database. It assumes schemas are shared between the two servers.

### DIFF
--- a/python/pipeline/utils/copydata.py
+++ b/python/pipeline/utils/copydata.py
@@ -1,0 +1,371 @@
+import numpy as np
+import datajoint as dj
+from getpass import getpass
+
+class DJServer:
+    """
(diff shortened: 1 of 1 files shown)


paulgfahey (approved):
Discussed with Cameron, implemented workaround for exporting data from one database to another behind firewall.  Due to datajoint restrictions to single database connection at a time, must fetch all data to RAM before inserting, so appropriate restrictions / looping commands are recommended.   Cameron confirmed testing was performed on his end before PR, and yields expected keys in table and expected analysis results across the two copies.

PR #451, opened 2022-11-22
Added support for new MotionCorrection methods

Neuro-CS:
Added new motion correction methods and modifications to supporting tables/code.

The structure is as follows:

shared.MotionCorrectionMethod + meso.MotionMethodForScan -> meso.MotionCorrection -> meso.MotionCorrection.MotionMethodUsed

Method 8 is currently suggested as the best method and is automatically filled out in MotionMethodForScan once RasterCorrection is done populating. This ensures users do not need to perform extra steps to process their scans. The method used is also recorded in a new part table, MotionMethodUsed. This is done because MotionCorrection cannot have MotionMethodForScan added as a dependency without recreating the table, so keys in MotionMethodForScan could be changed/deleted without it propagating downstream.

### DIFF
--- a/python/pipeline/meso.py
+++ b/python/pipeline/meso.py
@@ -5,6 +5,8 @@ import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
 import scanreader
+from scipy import ndimage
+from tqdm import tqdm
 
 from . import experiment, injection, notify, shared
 from .utils import galvo_corrections, signal, quality, mask_classification, performance
@@ -439,6 +441,9 @@ class RasterCorrection(dj.Computed):
         # Insert
         self.insert1(tuple_)
 
+        # Fill MotionMethodForScan table with default motion correction method
+        MotionMethodForScan.fill(key)
+
     def get_correct_raster(self):
         """ Returns a function to perform raster correction on the scan. """
         raster_phase = self.fetch1("raster_phase")
@@ -453,8 +458,21 @@ class RasterCorrection(dj.Computed):
 
 
 @schema
+class MotionMethodForScan(dj.Manual):
+    definition = """ # defines which motion correction method to use
+    -> ScanInfo
+    ---
+    -> shared.MotionCorrectionMethod
+    """
+
+    def fill(self, key, motion_correction_method=8):
+        self.insert1({**key, "motion_correction_method": motion_correction_method}, 
+                      ignore_extra_fields=True, skip_duplicates=True,)
+
+
+@schema
 class MotionCorrection(dj.Computed):
-    definition = """ # motion correction for galvo scans
+    definition = """ # experimental motion correction for galvo scans
 
     -> RasterCorrection
     ---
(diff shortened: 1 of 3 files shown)


python/pipeline/shared.py
```diff
@@ -188,6 +188,27 @@ class SurfaceMethod(dj.Lookup):
...
+
+@schema
+class MotionCorrectionMethod(dj.Lookup):
+    definition = """ # methods for motion correction
+    motion_correction_method    : tinyint unsigned
+    ---
+    correction_method_details   : varchar(1024)
+    """
+    contents = [
+        [0, "Legacy motion correction (before 2023). Code has been replaced."],
```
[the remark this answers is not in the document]
paulgfahey: am I correct in reading this as the code in this PR is no longer able to run method zero in the old, inefficient way?  Is it necessary to remove this capability?

python/pipeline/meso.py
```diff
@@ -519,55 +565,301 @@ def make(self, key):
             "fill_fraction": fill_fraction,
             "template": template,
         }
+
+        # Determine border cutoff to avoid edge artifacts
+        px_height, px_width = (ScanInfo.Field & key).fetch1("px_height", "px_width")
+        skip_rows = int(round(px_height * 0.10))
+        skip_cols = int(round(px_width * 0.10))
+
+        # Determine xy shifts via phase correlation
```
paulgfahey, line 574: could you add to this comment an explanation for why the field id and channel id variables have dummy entries?

python/pipeline/meso.py
```diff
@@ -519,55 +565,301 @@ def make(self, key):
...
+
+        # Run second iteration of motion correction using a global template
+        if key["motion_correction_method"] in (
+            5,
+            7,
+        ):
+
+            print("Running second iteration...")
+
+            scan = scan[0, :, :, 0, :]
```
paulgfahey, line 605: could you add a comment to explain this indexing?

python/pipeline/meso.py
```diff
@@ -519,55 +565,301 @@ def make(self, key):
...
+            scan = np.expand_dims(scan, axis=(0, 3))
+            template = galvo_corrections.create_refined_template(scan, x_shifts, y_shifts, key)
+
+            f = performance.parallel_motion_shifts  # function to map
+            kwargs = {
+                "raster_phase": raster_phase,
+                "fill_fraction": fill_fraction,
+                "template": template,
+            }
+            results = performance.map_frames(
```
paulgfahey, line 660: same as above, could you add a comment on why the field and channel are dummy inputs?

python/pipeline/meso.py
```diff
@@ -519,55 +565,301 @@ def make(self, key):
...
+            second_x_shifts, second_y_shifts, _ = galvo_corrections.process_xy_motion(
+                second_x_shifts,
+                second_y_shifts,
+                key,
+            )
+
+            x_shifts = second_x_shifts + old_x_shifts
+            y_shifts = second_y_shifts + old_y_shifts
+
+        # Run a second iteration of motion correction using multiple local templates
```
paulgfahey, line 685: Is it possible to consolidate this code with the above code, since it seems to be heavily redundant?

python/pipeline/meso.py
```diff
@@ -519,55 +565,301 @@ def make(self, key):
...
+                    smoothing=True,
+                )
+
+                # scan variable had excess fields and channels stripped away to reduce memory footprint
+                kwargs = {
+                    "raster_phase": raster_phase,
+                    "fill_fraction": fill_fraction,
+                    "template": local_template,
+                }
+                # Determine xy shifts via phase correlation
```
paulgfahey, line 782: same as above, please add comment on dummy field and channel variables.

python/pipeline/meso.py
```diff
@@ -519,55 +565,301 @@ def make(self, key):
...
         tuple_["motion_template"] = template
         tuple_["y_shifts"] = y_shifts
         tuple_["x_shifts"] = x_shifts
         tuple_["outlier_frames"] = outliers
         tuple_["y_std"] = np.std(y_shifts)
         tuple_["x_std"] = np.std(x_shifts)
 
+        # Create record of motion correction method
+        method_used_key = key.copy()
+        method_used_key.pop("channel")
```
paulgfahey, line 845: why is the channel not in the MotionMethodUsed?  won't that cause confusion if correction is performed on more than one channel?  Or are all these tables designed to bottleneck to a single channel at this point?

python/pipeline/utils/galvo_corrections.py
```diff
@@ -313,4 +315,277 @@ def correct_motion(scan, x_shifts, y_shifts, in_place=True):
...
+    Required to make template functions below pipeline agnostic.
+    
+    Returns:
+        tuple: (height, width, height_microns_per_pixel, width_microns_per_pixel)
+    """
+    # Define virtual modules here to prevent circular imports
+    reso = dj.create_virtual_module("reso", "pipeline_reso")
+    meso = dj.create_virtual_module("meso", "pipeline_meso")
+
+    # Fetch relevant field size info
```
paulgfahey, line 332: technically these if statements will fail if the key is somehow present in both pipelines.  Worth creating a pipeline exception to error out if key is in both?

python/pipeline/utils/galvo_corrections.py
```diff
@@ -313,4 +315,277 @@ def correct_motion(scan, x_shifts, y_shifts, in_place=True):
...
+def _get_pipe(key):
+    """
+    Small utility function to return instance of meso/reso pipeline.
+    Required to make template functions below pipeline agnostic.
+    """
+    # Define virtual modules here to prevent circular imports
+    reso = dj.create_virtual_module("reso", "pipeline_reso")
+    meso = dj.create_virtual_module("meso", "pipeline_meso")
+    
+    if len(reso.ScanInfo & key) > 0:
```
paulgfahey, line 355: same as above, add exception if key in both tables?

python/pipeline/shared.py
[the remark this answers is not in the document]
paulgfahey: 0 is old code, 1 is same method but new code, 2 is the low memory method?

python/pipeline/utils/galvo_corrections.py
```diff
@@ -313,4 +315,277 @@ def correct_motion(scan, x_shifts, y_shifts, in_place=True):
...
+    else:
+        raise PipelineException(f"The create_template() function does not currently support motion_correction_method {key['motion_correction_method']}")
+    del mini_scan
+    # * Anscombe tranform to normalize noise, increase contrast and decrease outliers' leverage
+    # ** Small amount of gaussian smoothing to get rid of high frequency noise
+
+    return template
+
+
+def create_refined_template(
```
paulgfahey, line 464: Consider consolidating with create_template with x_shifts/y_shifts as optional arguments with default values = None, case statements to use if non-None inputs are provided?

paulgfahey (changes requested):

python/pipeline/meso.py
```diff
@@ -439,6 +441,9 @@ def make(self, key):
         # Insert
         self.insert1(tuple_)
 
+        # Fill MotionMethodForScan table with default motion correction method
+        MotionMethodForScan.fill(key)
```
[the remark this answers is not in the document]
Neuro-CS: This has been changed to method 1, keeping it consistent with the current pipeline.

python/pipeline/utils/galvo_corrections.py
Neuro-CS: After going through it, there are enough cases that are different between the two that the logic of the merged function would be confusing to follow. Future methods may require new but different processing for the first and second template, so it seems better to keep them separate for now.

paulgfahey (commented):
Reviewed with Cameron.

PR #452, opened 2023-03-30
Fixing critical timing bug in clocktools

Neuro-CS:
Found a critical bug that caused errors when fetching timing/interpolation data for "fluorescence-*" or "deconvolution-*" settings.

When including a `unit_id` or `mask_id` in the key for convert_clocks, the internal function `fetch_timing_data` would pull from the ms_delay stored in `pipe.ScanSet.UnitInfo` and add it to the current timestamps. However, timestamps are stored in seconds while `ms_delay` was converted to milliseconds. `ms_delay` was previously not divided by 1,000 before being added to the current timestamps, leading to reported times that can be 1s - 20s off. This would also make units that are fairly synchronized appear completely uncorrelated.

When including just `field` in the key for convert_clocks, the field delay was calculated from the delay image stored in `pipe.ScanInfo.Field`. There was a bug where this was using the average value of the delay image, but delay images are based off of the first pixel recorded at the top depth. That means when clocktools later sliced the timestamps unique for the field's depth, extra delay was added. This delay was on the order of milliseconds (20ms - 200ms), so it should be less consequential than the unit_id issue mentioned above.

### DIFF
--- a/python/pipeline/utils/clocktools.py
+++ b/python/pipeline/utils/clocktools.py
@@ -173,6 +173,13 @@ def fetch_timing_data(
             raise PipelineException(msg)
 
         ## Determine field offset to slice times later on and set ms_delay to field average
+        ## Because field delay image uses top depth as 0, we need to subtract the minimum value of the first field's
+        ## delay_im that was recorded at that depth. This will give us the average delay for that field.
+        ##
+        ## WARNING: This requires an equality restriction on z, which is technically a float in the database. It
+        ##          is possible for this to fail for extreme values of z. At time of writing this only applies to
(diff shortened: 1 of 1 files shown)


paulgfahey (approved):
Discussed with Cameron, bug resulted from inconsistent storage of delay (milliseconds in some places and seconds in others), as well as the delay reference to the first pixel of the first roi per frame instead of first roi per depth.  Clocktools package was not widely used in atlab infrastructure at the time of bug detection, although individual users who adapted it might have been affected.  Slack channel broadcast at the time notified users, did not hear back from anyone affected in atlab.

PR #458, opened 2024-06-11
GPU motion correction

fdeguire03:
Add GPU awareness to motion correction function in galvo_corrections. Tried to maintain as much consistency with legacy CPU function as possible. Marginal additional optimizations likely possible if we were to rewrite entire function from scratch.

In simple testing, GPU-based code is about 1.5x faster than CPU-based code on 240x240 frame size.

Also added check at the top of galvo_corrections to import tukey method from the correct scipy submodule (location changed after scipy version 1.0.0).


fdeguire03:
Made by mistake, meant to go in fork for now.

PR #461, opened 2025-09-17
feat: add spike_method: 6 to reso

spapa013:
# Summary
Adds spike method 6 to pipeline/reso.py, ported from pipeline/meso.py.

# What changes

Method 6 uses CaImAn’s cmn.deconvolve_detrended, which detrends fluorescence traces prior to deconvolution.

The previous reso method (method 5) uses cmn.deconvolve (no detrending).

# Implementation notes

cmn.deconvolve_detrended requires the scan frame rate (fps). We fetch it from ScanInfo and bind it via functools.partial:

scan_fps = (ScanInfo & key).fetch1('fps')
deconvolve = partial(cmn.deconvolve_detrended, scan_fps=scan_fps)

The resulting deconvolve callable replaces the cmn.deconvolve used in method 5; all downstream steps are unchanged.

# Compatibility

No new dependencies on meso; the implementation is self-contained and fully compatible with the reso pipeline.

Method 5 remains available; this PR only adds method 6.

# Rationale

Using detrended deconvolution can reduce baseline drift effects before spike inference, improving robustness on slowly varying signals.

### DIFF
--- a/python/pipeline/reso.py
+++ b/python/pipeline/reso.py
@@ -1872,6 +1872,26 @@ class Activity(dj.Computed):
                 Activity.Trace().insert1({**key, 'unit_id': unit_id, 'trace': spike_trace})
                 Activity.ARCoefficients().insert1({**key, 'unit_id': unit_id, 'g': ar_coeffs},
                                                   ignore_extra_fields=True)
+                
+        elif key["spike_method"] == 6:  # dnmf
+            from pipeline.utils import caiman_interface as cmn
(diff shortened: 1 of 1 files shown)


paulgfahey:
Could you leave a brief description on the difference between methods 5 and six, that this code is a direct port from meso,  and check that the difference is fully encapsulated in the deconvolve_detrended function call and not distributed in multiple locations in meso?

paulgfahey (approved):
See comments.

